PR #71, opened 2021-11-15
UMAP enhancements

jamesb93:
The "iterations" (epoch) of UMAP was resetting to 0 when it organically completed its iterations. Really, it should stay at the cap to demonstrate to the user that it has finished the iterations.


jamesb93:
Hi @tremblap and @tedmoore can I request your eyes on this?

You can find the most updated version here:

https://deploy-preview-71--flucomalearn.netlify.app/reference/umap

PR #72, opened 2021-11-16
[Content] Buffer Explanation

jamesb93:
This PR is still WIP.

I aim to compose it such that it explains the rationale for buffers a bit better as well as provide a series of vignettes of objects / helpers for dealing with them.﻿


jamesb93:
@weefuzzy @tremblap @tedmoore i'd be interested in hearing your thoughts on different buffer interface things to deal with.

My ideas are...

1. Buffer to list in Max, less of an issue in pd/sc but I don't know the edges of the latter two
2. Iterating through buffers with poke/peek~
3. Reshaping with select/selectevery~ and bufcompose~
4. Storing to/from disk as a state management thing
5. "Playing data" ala Timo Hoogland generating wavetables with MLPClassifier

Basically just a brainstorm of buffers :)

jamesb93:
P.S you can see the stuff I've already written:

https://deploy-preview-72--flucomalearn.netlify.app/overviews/buffers

tremblap:
is this still work in progress? Vercel fails but I can write my views on your questions:

> Buffer to list in Max, less of an issue in pd/sc but I don't know the edges of the latter two

In Pd one uses `array get` to get the array :) it is quite powerful and it is accessible within expr so a very good interface choice

SC: special note needed to remind people that we cannot get via UDP any length. it is in getToFloatArray helpfile

Max: we need to remind people of the zl 256 default limit

> Iterating through buffers with poke/peek~

PD: `array get` and `array set` enable full dump/write or partial dump/write so we're set there too

> Reshaping with select/selectevery~ and bufcompose~

PD: these are useful, as well as flatten, to go around the fact that multiarray is a cheat around the fact that there are no multichannel buffers - which reminds me we need a special note

> Storing to/from disk as a state management thing

yes - and cross CCE and cli compatible and python, so a good bridge

> "Playing data" ala Timo Hoogland generating wavetables with MLPClassifier

wavetable synth you mean? yep!

> Basically just a brainstorm of buffers :)

there is also the fact that it is fast to address so the control(language)/audio(server) boundary is crossable (semi) elegantly

my 2c for now

jamesb93:
> is this still work in progress? Vercel fails but I can write my views on your questions:

It was WIP but is, I'd say, stale now so I'll have to give it some TLC at some point. Your pointers are helpful though. Many little edges and corners to remember everywhere!

PR #75, opened 2021-11-18
KDTree Reference

jamesb93:
WIP. This adds a reference for the KDTree with an interactive example.


jamesb93:
Still needs a tie in to the upcoming 2D Plotting Videos. These are imminently slotted to release.

@weefuzzy @tedmoore @tremblap what more can we do to bolster the explanation of the kdtree?

PR #76, opened 2021-11-22
Ampslice Interactive Explanation

jamesb93:
This adds basic functionality for an AmpSLice demonstration with the two feature curves overlapping.

It has some warts, in so far as the feature curves don't really draw in a sane way. The curves update at the edges + I'm smoothing the whole window, when I really should smooth chunks and piece it all back together in a buffer bigger than the view... or something. :shrug:﻿


jamesb93:
Drawing vertical lines on the points where there the fast exceeds the slow envelope with a given (controllable??) threshold would be good too.

Follow up with some fancy graphics to explain things perhaps...

jamesb93:
Currently not building because of ESM6 sadness:

https://github.com/danchitnis/webgl-plot/issues/68

PR #82, opened 2021-11-26
[Content] Batch processing overview

jamesb93:
This adds a small overview on batch ways of approaching batch processing. It is a little shallow in that it doesn't discuss _all_ the knottier parts of each CCE. Instead I offer a few vignettes maintaining parity over each CCE.


weefuzzy (changes requested):
There *is* a distinction to be drawn in the SC case though. If you used `processBlocking` in the first case, then you can make assumptions about the order of completion that allow you to write instructions in the order you want them carried out and have `s.sync` signal completion. Using `wait` serially in non-blocking mode is actually less efficient because it causes extra traffic between synth and language: the trick in using blocking 0 effectively in SC is to partition work into what can be truly done in parallel and to minimize language side waits (e.g. a bunch of different descriptors could run on the same source in parallel; you'll then want to synchronise on the condition that they're all done before marshalling into a dataset)

jamesb93:
> There _is_ a distinction to be drawn in the SC case though. If you used `processBlocking` in the first case, then you can make assumptions about the order of completion that allow you to write instructions in the order you want them carried out and have `s.sync` signal completion. Using `wait` serially in non-blocking mode is actually less efficient because it causes extra traffic between synth and language: the trick in using blocking 0 effectively in SC is to partition work into what can be truly done in parallel and to minimize language side waits (e.g. a bunch of different descriptors could run on the same source in parallel; you'll then want to synchronise on the condition that they're all done before marshalling into a dataset)

Hmm okay. So:

```js
FluidBufHPSS.processBlocking(s, source:src, harmonic:h, percussive:p, numChans:1);
FluidBufCompose.processBlocking(s, source:h, destination:result, destStartChan:0);
FluidBufCompose.processBlocking(s, source:p, destination:result, destStartChan:1);
```

is actually how one might do it to avoid using `wait` at the expensive of "yellowing" the server? Then `.process()` is the nonblocking version that requires one to think about awaiting the return of things and synchronicity of the server within and around a `Routine` type structure?

weefuzzy:
> is actually how one might do it to avoid using `wait` at the expensive of "yellowing" the server? 

Yes 

> Then `.process()` is the nonblocking version that requires one to think about awaiting the return of things and synchronicity of the server within and around a `Routine` type structure?

No, the need for `Routine`s applies in both cases if you want to write sequential code that involves waiting of any form. The distinction is that, in the blocking 0 case, sequential code doesn't help with much because `s.sync` is no longer informative. 

so 

```js
fork {
    FluidBufHPSS.processBlocking(s, source:src, harmonic:h, percussive:p, numChans:1);
    FluidBufCompose.processBlocking(s, source:h, destination:result, destStartChan:0);
    FluidBufCompose.processBlocking(s, source:p, destination:result, destStartChan:1);
    s.sync;
}
```

is pretty easy to follow in terms of the order things happen in (as written), and being able to control when the routine finishes (after `sync`) 

Whereas: 
```js
fork {
    FluidBufHPSS.process(s, source:src, harmonic:h, percussive:p, numChans:1);
    FluidBufCompose.process(s, source:h, destination:result, destStartChan:0);
    FluidBufCompose.process(s, source:p, destination:result, destStartChan:1);
    s.sync;
}
```
would be surprising (by the `Routine` returning  almost immediately, but also dependencies between tasks not being respected). In that scenario you either need to `.wait`, or use `actions` to control flow.

jamesb93:
> > is actually how one might do it to avoid using `wait` at the expensive of "yellowing" the server?
> 
> Yes
> 
> > Then `.process()` is the nonblocking version that requires one to think about awaiting the return of things and synchronicity of the server within and around a `Routine` type structure?
> 
> No, the need for `Routine`s applies in both cases if you want to write sequential code that involves waiting of any form. The distinction is that, in the blocking 0 case, sequential code doesn't help with much because `s.sync` is no longer informative.
> 
> so
> 
> ```js
> fork {
>     FluidBufHPSS.processBlocking(s, source:src, harmonic:h, percussive:p, numChans:1);
>     FluidBufCompose.processBlocking(s, source:h, destination:result, destStartChan:0);
>     FluidBufCompose.processBlocking(s, source:p, destination:result, destStartChan:1);
>     s.sync;
> }
> ```
> 
> is pretty easy to follow in terms of the order things happen in (as written), and being able to control when the routine finishes (after `sync`)
> 
> Whereas:
> 
> ```js
> fork {
>     FluidBufHPSS.process(s, source:src, harmonic:h, percussive:p, numChans:1);
>     FluidBufCompose.process(s, source:h, destination:result, destStartChan:0);
>     FluidBufCompose.process(s, source:p, destination:result, destStartChan:1);
>     s.sync;
> }
> ```
> 
> would be surprising (by the `Routine` returning almost immediately, but also dependencies between tasks not being respected). In that scenario you either need to `.wait`, or use `actions` to control flow.

Ahhhh... this makes sense :) Okay, clearly need to develop out some .scd scripts to nuance what has been said here. It's also probably relevant to even link directly to this PR :) Thanks @weefuzzy

jamesb93:
Okay so I've had a crack at this [here](https://github.com/flucoma/learn-website/pull/82/commits/72bd915d83daeab0ffbe283dbc54f43bf8bdd4dd) and works (to a point). I'm unsure how to await the return of the entire loop `files.do` loop as the print that happens "afterwards" comes immediately due to the async nature of it all. Should I be wrapping the loop in its own `Routine/Fork` and then syncing after that or something?

jamesb93:
Okay I've dived into some more SC code though I am still feeling out of my depth. I've managed to somewhat solidify the difference in my head for how the threading model works and how batch processing can (potentially) be made more efficient. 

I need two things to finish this off I think.

1. A more rigorous and knowledgeable set of eyes on the SuperCollider code. @tedmoore and @weefuzzy if you have time at some point that would be grand.
2. Some help fleshing out the discussion of batch processing in SC. The benefits of mixing `@blocking 0` into a batch process in Max are obvious to me but for SC I'm not sure how to properly express how the change to using .process() can be of benefit for a workflow.

weefuzzy:
> I need two things to finish this off I think.
> 
>     1. A more rigorous and knowledgeable set of eyes on the SuperCollider code. @tedmoore and @weefuzzy if you have time at some point that would be grand.

From a quick eyeball, it looks basically ok, although the threaded example is 🙀: but that just underscores that we need a better way to abstract away some of the boilerplate and horrible state management for that scenario in general. Some of the comments are a bit confusing, and I think we can make the code a wee bit leaner, so I'll attack it and push an update when I can (unless @tedmoore gets there first). 

>     2. Some help fleshing out the discussion of batch processing in SC. The benefits of mixing `@blocking 0` into a batch process in Max are obvious to me but for SC I'm not sure how to properly express how the change to using .process() can be of benefit for a workflow.

The tradeoffs are basically the same in all environments, but it's easier to get into trouble in sclang by launching loads and loads of threads. 

* In all environments, non-blocking disrupts normal control flow. There's (small) precedent in Max with the `stretch~` object and other 3rd party externals that use worker threads (e.g `mubu.process`). 
* Meanwhile, SC users are basically used to asynchrony because of the split processes. However, non-blocking adds a further layer (which is a brain twister) and we don't have synchronisation or queuing methods on the server, so when there's many small jobs it's actually easy to make things much slower due to added language-server traffic 
* In any environment, adding threads isn't a simple gateway to faster batch processing: once you add too many, then the scheduling overhead starts to kill performance.

static/examples/batch/batch-decomposition-threads.scd
```diff
@@ -0,0 +1,74 @@
...
+    // Maintain a counter for the number of files processed.
+    // This will be use to determine when all of the files are finished working.
+    var counter = 0;
+
+    // Remove any files from the audioFiles collection so that only those with a .wav extension remain
+    var wav = audioFiles.files.removeAllSuchThat({ |file| file.extension == "wav" });
+
+    // In the language iterate over all the files in the example audio files directory
+    wav.do({ |file|
+        fork ({
```
weefuzzy, line 21: This will potentially start a lot of worker threads on the server, so it could actually be quite slow. There isn't a nice solution to this yet. The safest / most paternalistic would be to have a FluCoMa-wide thread pool to cap the number of OS threads in flight for async jobs, but that's decidedly non-trivial. Next best is to cap from the language, but that's not hugely simple either (I guess using a `Semaphore` might do it)

static/examples/batch/batch-decomposition.scd
```diff
@@ -0,0 +1,42 @@
+(
+s.waitForBoot{
+    Task{
```
weefuzzy, line 3: We should be consistent between using `fork` / `Task` / <whatever> between examples

static/examples/batch/batch-slicing-threads.scd
```diff
@@ -0,0 +1,86 @@
...
+    var data = FluidDataSet(s); // A FluidDataSet() for storing the analysis of each slice (mean centroid)
+
+    // Create a condition that hangs while processing is still happening and unhangs when its complete.
+    var processingCompleted = Condition();
+    var counter = 0; // Also keep track of how many slices we have processed.
+    s.sync;
+
+    // Slice the drum Buffer() based on consecutive changes in spectral frames
+    // The samples at which slices are detected will be written into the "slices" Buffer()
+    FluidBufOnsetSlice.processBlocking(s, drum, indices:slices, metric:9,threshold:0.5);
```
weefuzzy, line 15: `s.sync` should go after this so that we know `slices` is actually populated before `loadToFloatArray`

static/examples/batch/batch-slicing-threads.scd
```diff
@@ -0,0 +1,86 @@
...
+                // Then calculate the statistical analyses of the centroid for each segment.
+                // This will return write the output into a new Buffer() called sliceStats that will be 1 channel and 7 frames in size.
+                FluidBufStats.process(s, source:sliceCentroid, stats:sliceStats).wait;
+
+                // Then extract only the first frame from the 7 frames and write it into a Buffer() called sliceMean.
+                // This isolates a singular value, the average spectral centroid for the segment.
+                FluidBufSelect.process(s, source:sliceStats, destination:sliceMean, indices:0).wait;
+
+                // Now that we have a 1x1 Buffer() holding the mean centroid for that segment we can add it to the FluidDataSet.
+                data.addPoint(sliceIndex, sliceMean);
```
weefuzzy, line 61: Preference for `setPoint` here

static/examples/batch/batch-slicing-threads.scd
```diff
@@ -0,0 +1,86 @@
...
+                s.sync;
+                counter = counter + 1;
```
weefuzzy, line 63: We could just track the `size` of the DataSet instead (which implies a further server call, but is perhaps clearer in terms of expressing intent)

static/examples/batch/batch-slicing.scd
```diff
@@ -0,0 +1,68 @@
...
+    fork({
+        var drum = Buffer.read(s,File.realpath(FluidDataSet.class.filenameSymbol).dirname.withTrailingSlash ++ "../AudioFiles/Nicol-LoopE-M.wav");
+        var slices = Buffer(s); // The slicer (FluidBufOnsetSlice) will write into this buffer the samples at which slices are detected
+        var sliceAnalysis = Buffer(s); // A Buffer() for writing the analysis from FluidSpectralShape into
+        var sliceCentroid = Buffer(s); // A Buffer() for extracting the centroid into
+        var sliceStats = Buffer(s, 7); // A Buffer() for writing the statistical analyses of the centroid into
+        var sliceMean = Buffer(s, 1); // A Buffer() for extracting just the mean from the 7 statistical analyses.
+
+        var data = FluidDataSet(s); // A FluidDataSet() for storing the analysis of each slice (mean centroid)
+        s.sync;
```
weefuzzy, line 13: Don't think this is needed because of the `wait` below

static/examples/batch/batch-slicing-threads.scd
weefuzzy: Not sure why we have `processBlocking` for this example (which is showcasing async approach) and `process` in the other (which is showcasing sync approach)

static/examples/batch/batch-slicing-threads.scd
```diff
@@ -0,0 +1,86 @@
...
+                // We extract only the first channel of the 7 channels (the centroid) and put it into a Buffer() named sliceCentroid
+                FluidBufSelect.process(s, source:sliceAnalysis, destination:sliceCentroid, channels:0).wait;
+
+                // Then calculate the statistical analyses of the centroid for each segment.
+                // This will return write the output into a new Buffer() called sliceStats that will be 1 channel and 7 frames in size.
+                FluidBufStats.process(s, source:sliceCentroid, stats:sliceStats).wait;
+
+                // Then extract only the first frame from the 7 frames and write it into a Buffer() called sliceMean.
+                // This isolates a singular value, the average spectral centroid for the segment.
+                FluidBufSelect.process(s, source:sliceStats, destination:sliceMean, indices:0).wait;
```
weefuzzy, line 58: Does an essentially serial chain of work really demonstrate why we'd want to have worker threads? I guess extracting two different features would allow you to exploit some parallelism instead, but might just be more confusing.

static/examples/batch/batch-slicing-threads.scd
```diff
@@ -0,0 +1,86 @@
...
+            [0,1]
+            [1,2]
+            [2,3]
+            [3,4]
+
+            This will give us the left and right boundary of each segment
+            We can use this to inform the analysis (FluidBufSpectralShape) where to start analysing and how many frames to analyse
+            */
+
+            fork ({
```
weefuzzy, line 33: As with the other async example, this potentially starts a lot of threads on the server, which may be sub-optimal. 

As a side note, processing lots of very small slices isn't actually a great use case for threading as currently implemented, because we have to copy the whole source (rather than just the slice) into the new context every time. So, for a large source with lots of slices to process, it can end up being considerably slower than just ploughing through one by one. This is due to be fixed, but requires some internal redesign

static/examples/batch/batch-slicing.scd
```diff
@@ -0,0 +1,68 @@
...
+
+        // Slice the drum Buffer() based on consecutive changes in spectral frames
+        // The samples at which slices are detected will be written into the "slices" Buffer()
+        FluidBufOnsetSlice.process(s, drum, indices:slices, metric:9,threshold:0.5).wait;
```
weefuzzy, line 17: Perhaps use processBlocking here,l to be consistent?

static/examples/batch/batch-slicing.scd
```diff
@@ -0,0 +1,68 @@
...
+                // Then calculate the statistical analyses of the centroid for each segment.
+                // This will return write the output into a new Buffer() called sliceStats that will be 1 channel and 7 frames in size.
+                FluidBufStats.processBlocking(s, source:sliceCentroid, stats:sliceStats);
+
+                // Then extract only the first frame from the 7 frames and write it into a Buffer() called sliceMean.
+                // This isolates a singular value, the average spectral centroid for the segment.
+                FluidBufSelect.processBlocking(s, source:sliceStats, destination:sliceMean, indices:0);
+
+                // Now that we have a 1x1 Buffer() holding the mean centroid for that segment we can add it to the FluidDataSet.
+                data.addPoint(slice_index, sliceMean);
```
weefuzzy, line 59: prefer `setPoint`

static/examples/batch/batch-decomposition-threads.scd
```diff
@@ -0,0 +1,74 @@
...
+
+    // Create a condition that hangs while processing is still happening and unhangs when its complete.
+    var processingCompleted = Condition();
+
+    // Maintain a counter for the number of files processed.
+    // This will be use to determine when all of the files are finished working.
+    var counter = 0;
+
+    // Remove any files from the audioFiles collection so that only those with a .wav extension remain
+    var wav = audioFiles.files.removeAllSuchThat({ |file| file.extension == "wav" });
```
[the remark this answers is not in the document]
jamesb93: so much nicer... thanks

static/examples/batch/batch-decomposition-threads.scd
jamesb93: Yeah this is only pleasant with these small jobs...

I'll looking into the `Semaphore` and update accordingly.

static/examples/batch/batch-decomposition.scd
jamesb93: Easy fix, thanks for pointing out. There are a few places where I should make the style consistent too. I have a preference for ({}) whenever the function is not anonymous. So basically everywhere except `someThing(action:{})`

static/examples/batch/batch-slicing-threads.scd
jamesb93: Because I am bad at copy and paste :)

Do you think it is worth using a Semaphore here too? it's entirely possible to want to do a butt load of slices and maybe its good to show that you may need to cap the concurrent jobs on the server?

static/examples/batch/batch-decomposition-threads.scd
```diff
@@ -14,17 +14,20 @@ s.waitForBoot({
...
-    var wav = audioFiles.files.removeAllSuchThat({ |file| file.extension == "wav" });
+    var wav = audioFiles.files.select({ |file| file.extension == "wav" });
 
+    var pool = Semaphore(4); // Create a Semaphore that allows 4 consecutive jobs
     // In the language iterate over all the files in the example audio files directory
     wav.do({ |file|
         fork ({
-            var src = Buffer.read(s, file.fullPath.asString); // Create a buffer with the source audio in a buffer called "src"
+            var src, h, p, result;
+            pool.wait; // We tell the Semaphore above that we are going to be doing some work and one of the 4 workers needs to WAIT.
```
weefuzzy, line 24: Comment isn't quite accurate, I think: we're (in effect) queueing up this portion of work via the semaphore. If there are 4 jobs already underway, then this will wait.

static/examples/batch/batch-slicing-threads.scd
jamesb93: I've put in a semaphore :)

static/examples/batch/batch-slicing.scd
jamesb93: Seems to be okay without it.

static/examples/batch/batch-slicing-threads.scd
jamesb93: Hmm... with a semaphore doesn't it show that you could process multiple segments simultaneously?

static/examples/batch/batch-slicing.scd
jamesb93: I do in practice but find addPoint is more robust at throwing meaningful errors when you overwrite your data (potentially). My belief is there is pedagogical value in showing add first and set as the more responsibility, less code.

static/examples/batch/batch-decomposition-threads.scd
jamesb93: Nuanced that bad boy

static/examples/batch/batch-slicing-threads.scd
weefuzzy: Yes, each chain of `spectralshape` ▶️ `select` ▶️ `stats` ▶️ `select` will launch independently. I guess my concern is that by then enforcing a sequential flow within the chain, we're muddying the waters a bit because there's even more to think about. 

Do we really need the `select`s here – pedagogically speaking? Is there an argument for a simpler (if more redundant) process of 

`FluidSpectralShape.process(...args..., action: {FluidBufStats.process(...,action:{pool.signal})`

So that we so a 'familiar-in-SC' way to do deal with the asynchrony (callbacks), and minimize the number of steps, so that the basic point is emphasised

weefuzzy:
One more thing to think about is which `Buffer`s we can `free` (best practices and all that). In the `DataSet` examples, we can presumably free all the `Buffers` for that sub-task as soon as the point is added. For the `Dictionary` ones, we'll need to preserve the `result` `Buffer` , but the temporaries should be `free`ed (which doesn't happen when we exit the scope).

static/examples/batch/batch-slicing-threads.scd
jamesb93: Callbacks make me sad because the code becomes indented really fast and for me generally kills readability. I don't really know what SC people prefer and I look toward languages I know better like `js` where callbacks are being phased out in favour of .() chaining with promises (just as a general trend I've observed and read about).

static/examples/batch/batch-slicing-threads.scd
jamesb93: It's not a great case but I think it is good to show the ergonomics of it at least and to have parity with what exists as a Max patch too. I would opt to keep it there but add more information in the comments about the potential downsides due to copying etc.

static/examples/batch/batch-slicing.scd
jamesb93: Done

jamesb93:
> One more thing to think about is which `Buffer`s we can `free` (best practices and all that). In the `DataSet` examples, we can presumably free all the `Buffers` for that sub-task as soon as the point is added. For the `Dictionary` ones, we'll need to preserve the `result` `Buffer` , but the temporaries should be `free`ed (which doesn't happen when we exit the scope).

I need to go through all of this and check it. I was stupidly under the impression that because it was scoped to the function the server would would also know about when to free em up 🙈

static/examples/batch/batch-slicing-threads.scd
jamesb93: I like that over tracking with a counter.

static/examples/batch/batch-slicing-threads.scd
weefuzzy: I prefer futures / promises and continuations too, but we have what we have – and I think there's an awful lot to be said for working with the idiomatic norms of a language, especially for didactic cases like this, in terms of getting the point across: if people aren't used to a `wait`-like idiom (which is less common in SC) then that's an extra chunk of thinking that needs to be done in order to understand what's going on. 

Excessive Indentation can be alleviated by moving callbacks into variables 
```js
var onSpectralShapeComplete = { FluidBufStats.process(..., action: onStatsComplete}; 
var onStatsComplete = {//write result & signal pool}; 
FluidBufSpectralShape.process(...,action:onSpectralShapeComplete}; 
```

This has the downside of de-localising the flow, though. 

Once we have our processes reporting failure back to the language (which we don't yet), then doing a continuation style idiom like `then()` would be more tractable, although SC folk might not like it...

static/examples/batch/batch-slicing-threads.scd
jamesb93: This seems to be problematic likely due to my coding... it always prints the size to the console which confuses what is going on. Is there a way to nab `data.size` without it hassling the user?

static/examples/batch/batch-slicing-threads.scd
jamesb93: Okay, I agree that removing one of those "what?" moments is valuable and so I'll callback-ify this one.

static/examples/batch/batch-slicing-threads.scd
weefuzzy: IIRC, the default action for `size` is a `postln`, so you'd want to override that to move into a local var, which perhaps means you can get rid of the sync and stuff

```js
data.addPoint(sliceIndex, sliceMean);
data.size(action: {|dataSetSize| 
  if(dataSetSize == wav.size, { processCompleted.unhang; }); 
}); 
```

static/examples/batch/batch-slicing-threads.scd
jamesb93: I have put callbacks in @weefuzzy but introduced more bugs 😂

1. Lots of warnings about freeing already free buffers (which I don't get)
2. Repeated sliceIndex printouts which shouldn't happen I don't think...
Care to take a look?

weefuzzy: Ah, yeah, I see. It's because the callbacks on the dataset (which is a singular object) overwrite each other. We've had fun with this before :sob:

So, with the `FluidBuf*.process` stuff, we're calling class methods, and these are essentially like a little factory that creates an object and has an action associated with that object, and then it all goes away again. However, with a `FluidDataSet`, we're dealing with a persistent instance, so when we register an action with a method then – as it stands – the mapping between the action and the method is 1:1, so that if we call the method again with a different action, then the original one is lost. It's like this because we lifted the action handling, basically, from `Buffer`, which exhibits the same behaviour. Obviously, what could be less confusing is to map actions to *invocations* somehow (i.e. I don't know how, offhand)

The solution here is not to do the cleanup in the `FluidDataset.size` action, but to hoist into the final `FluidBufSelect`. Push incoming

jamesb93:
@tremblap and @tedmoore would you be able to review this one formally now?

@tedmoore has already provided some feedback on the writing and the SC scripts but I think now it would be good to refamiliarise yourself with it and then we can convene for a once over together.

[Link](https://learn-git-fork-jamesb93-content-batch-processing-flucoma.vercel.app/overviews/batch-processing )

tedmoore:
The example code here is great and is going to be really useful to the reader. I think that should go right at or near the tip top of the page. As a reader, I (and others I think) would gravitate towards that.

Generally I think that the page buries the lead. it's quite far down the page that someone gets to the header: `Batch Processing` Which starts with "So how do we approach batch processing with the FluCoMa objects?" (In my estimation, this is actually the question that would bring someone to the page, so it should be the lead of the page.) In the words of George Clinton "Give the people what they want when they want and they won't sit all the time".

Overloading the `+` operator in SC doesn't actually appear in the demo scripts, so I don't think it's necessary to take a reader through it prior to them doing batch processing. I think the same is true for `[vexpr]`?

tremblap:
comments on the Pd batch process (apart that they are awesome! streamlined and clear and enough feedback and readable) - and the

- the symbol at the top could use ```media/Nicol-LoopE-M.wav``` as it will be found by a proper install by [soundfiler](although it doesn't seem to work in [file glob] so I've posted a question to the list.
- [v sr] could be used instead of [s sr] (a neat trick in Max too)
- the windows are all in the middle of a large screen - we should save the patches at the top left of the screen

meta comment: the vercel integration is amazing to review.
integration comment: I really like the tutorial/code/text integration here (@tedmoore that might be helpful for bufnmf in progress)

tremblap:
one more comment: the dynamic object creation in the pd batch processing is naughty. I think using our evil friend [struct] would be the canonical way. I'm happy to recode that bit if that helps?

jamesb93:
> comments on the Pd batch process (apart that they are awesome! streamlined and clear and enough feedback and readable)

Thanks :)

> * the symbol at the top could use `media/Nicol-LoopE-M.wav` as it will be found by a proper install by [soundfiler](although it doesn't seem to work in [file glob] so I've posted a question to the list.

That makes sense to me because the glob doesn't necessarily reference a path that pd will know about and can extend to the whole filesystem.

> * [v sr] could be used instead of [s sr] (a neat trick in Max too)

I am using a `v sr` in the playback subpatch but store the variable with a named send. Is that confusing?

> * the windows are all in the middle of a large screen - we should save the patches at the top left of the screen

That is weird, they open at top left for me?

jamesb93:
> one more comment: the dynamic object creation in the pd batch processing is naughty. I think using our evil friend [struct] would be the canonical way. I'm happy to recode that bit if that helps?
> 
> **EDIT** I'm not sure I agree with myself here. It works and it is similar. So let's leave it if everyone agrees and we can see what the Pd community says.

Just looking at struct makes me feel woozy. I think what I've done, albeit hacky, is simpler to follow. That said, it does seem hacky (I think thread safe though) so I'm not sure.

jamesb93:
> Generally I think that the page buries the lead. it's quite far down the page that someone gets to the header: `Batch Processing` Which starts with "So how do we approach batch processing with the FluCoMa objects?" (In my estimation, this is actually the question that would bring someone to the page, so it should be the lead of the page.) In the words of George Clinton "Give the people what they want when they want and they won't sit all the time".

I'm not sure what the people _want_ in this regard. I think that this overview is as much showing a solution to a problem as highlighting a longstanding issue in the CCEs that form a counterweight to the interface of the FluCoMa tools.  I think the preamble provides a necessary context, especially for people who may have an inkling there is _something wrong_ in Max and Pd😂 but can't necessarily put their finger on it (whereas SC solves these problems more elegantly and idiomatically). Maybe what the article needs is a better foreword and stronger introduction? "Batch processing corpora with the FluCoMa toolkit. Then, following your other comment about the location of the code, there could be a "for dummies" type link at the top of the page that says: "if you want to skip straight to the code, you can find it down here"?  

> Overloading the `+` operator in SC doesn't actually appear in the demo scripts, so I don't think it's necessary to take a reader through it prior to them doing batch processing. I think the same is true for `[vexpr]`?

Isn't the `+` operator overloaded by default? If you add together two integers or two `Array`s then + is overloaded and those classes will have something like an internal `__add__` method. Maybe what I've written suggests I'm going to show you how to overload them when what I mean is that the interface is consistent due to overloading that is extant?

jamesb93:
> The example code here is great and is going to be really useful to the reader. I think that should go right at or near the tip top of the page. As a reader, I (and others I think) would gravitate towards that.
> 
> Generally I think that the page buries the lead. it's quite far down the page that someone gets to the header: `Batch Processing` Which starts with "So how do we approach batch processing with the FluCoMa objects?" (In my estimation, this is actually the question that would bring someone to the page, so it should be the lead of the page.) In the words of George Clinton "Give the people what they want when they want and they won't sit all the time".
> 
> Overloading the `+` operator in SC doesn't actually appear in the demo scripts, so I don't think it's necessary to take a reader through it prior to them doing batch processing. I think the same is true for `[vexpr]`?

Okay, I've pushed a major reorientation of the content. I was discussing how to best order this in light of your feedback with my beloved who said: "You don't want this to be like a food blog where they spend 1000 words talking about the lineage of the recipe do you?", which made it click. You're right. I've compromised somewhere in the middle in reducing fluff and moving most of the discursive stuff to a section at the bottom which makes it more "optional". It definitely feels more punchy now, and takes you to the guts faster leaving the flowery stuff open to those who are interested.

@tedmoore what do you think?

jamesb93:
After a pow-wow with @tedmoore, I've finalised the changes and merged into main. Woohoo 🚀

PR #89, opened 2021-12-01
dataset reference entry

jamesb93:
Adds a reference entry for dataset object﻿


jamesb93:
@tedmoore this is on our learn platform mega plan. I've e-mailed you about the granularity of dataset knowledge but perhaps we can open a transparent dialogue here.

I think that simply pointing out the caveats and limitations of the dataset is enough for the learn platform and doesn't require an overview for several reasons.

1. I'm not sure how an overview would differentiate its content from the DataSet as much of the technical and abstract explanation can take place in the reference (describing association, identifiers, points, data).
2. The help files are very adequate already in showing the different ways that you can poke around in a dataset and making a web version of that seems like it would just be worse. 
3. Making examples on the web would be duplicating examples found in the CCE documentation and it seems counterproductive to just upload them to learn as well as being another point of failure for link management.

What are your thoughts?

PR #93, opened 2021-12-03
KMeans Reference

jamesb93:
- add ml5 as a dependency
- fix some naughties in DataSetView components
- basic kmeans vis


jamesb93:
Would be curious what you have to say about this one @tedmoore @weefuzzy . It has taken me a while to get to this point. Actually explaining Kmeans is quite rough and the sklearn resource does it well for experts but for beginners it is abstract.

weefuzzy (commented):
I think the language can be made simpler: we end up with jargon like 'cluster' and 'centroid' very quickly (whereas, I think part of the point might be to introduce the jargon carefully).

For instance, rather than 
> KMeans is an object that facilitates learning _clusters_ from data. Given a collection of points and a user-specified _number of clusters_,

we might say 

> The KMeans algorithm performs 'clustering', which means that–given some data–it tries to find a way to divide that data up into a number of groups (the 'clusters').  

(and so on)

The way some of its assumptions work could be demonstrated by using your neat simulator with less friendly data. Four gaussian groups with the same variance is pretty much optimal for KMeans. Some of the other examples from the sklearn docs will help illustrate what it leans towards and the sorts of things it will *never* cluster (e.g. concentric rings of different classes)

jamesb93:
There are some more datasets available. Should I perhaps make the UX more obvious for that? Or do you think what is there doesn't communicate the pitfalls well enough?

weefuzzy:
Ah, I didn't notice. However, I think maybe using a different instance of the widget, and having the page through what's going on for a few, quite tightly curated things would help (and certainly nudge it beyond the sklearn stuff, which assumes the reader is familiar enough to intuit why some stuff will work / not work)

jamesb93:
While this PR is closed I'm still thinking of ways in which to integrate your changes mentioned above @weefuzzy. It was more to get something on the page and to nuance later :)

PR #138, opened 2022-01-04
[Content] Median Filters Overview

jamesb93:
Where to look: `/overviews/median-filters`

- add median-filtering skeleton
- add images and video
- remove whitespace
- create a video component for mp4/webm
- make webm and mp4 binaries for git
- add assets


tremblap (commented):
great PR (I know where to look and what to look at!)

otherwise I recognise the content. I would probably weaken the blue down in the movie as it takes a lof of visual bandwidth but it is very clear what is happening and enough surprises in there to show what it does and what it doesn't.

tedmoore (changes requested):
Awesome. This will be great to link to for HPSS!

`is that they are good at rejecting short disruptions to a signal whilst doing a reasonable job of preserving the shape of the “good” bit` I'm not sure about `rejecting` here, i think `removing` is a better verb for what's going on? 

I do think that `good bit` should be changed or removed. it's removing outliers, but i don't think we know if they're "good" or "bad". sometimes we want the outliers. also for HPSS it's not that the harmonic or percussive parts that get median filtered out are good or bad, they're just the outliers.

`and the result of applying a median filter over this.` specify that it's the dotted red line.

`What is useful about median filters for our purposes here is that they are good at rejecting short disruptions to a signal, whilst doing a reasonable job of preserving the shape of the ‘good’ bit. For instance, here’s a sine wave that’s been disrupted by a series of spikes (blue line), and the result of running a median filter over this. We would call these spikes outliers.` this paragraph appears twice.

the video is very pretty!

`if we sorted the sample values in the window by their value, then the median is the sample value that ends up in the middle of this sorted list.` this sentence can be in the active voice. 

I would also add one note in here about how it gets used. You say that we are interested in the median value, but not that it is the median value of the window that is used for the output of the median filter.

tedmoore:
I say go for this and when it's merged, let me know so that I can link to it from the HPSS page which right now has an empty link! 😬

jamesb93:
> I say go for this and when it's merged, let me know so that I can link to it from the HPSS page which right now has an empty link! 😬

YOLO - we can go back and fix and make flashy later. Let's engage our inner content creators!

PR #139, opened 2022-01-04
[Content] Sinusoidal Modelling

jamesb93:
Where to look: `/overviews/sinusoidal-modelling`


tremblap (commented):
I love the text. The audio examples are not in the preview. I think the first graph is perfect, the 2nd is confusingly too dense and the 3rd one is only black and white (and I don't know why there is a diff in the PR that has a coloured voices which is clearer that doesn't appear on the preview)

so if we have sounds and the coloured version of v3 and maybe a slighly less vertically compressed of v2 then I think it is perfect.

tedmoore:
I also like this very much. And I agree with PA's assesmennts.

`The partial structure should be pretty clear to see from the spectrogram.` I think this needs to be reworded so that it is more clear that by `partial` you mean an upper frequency component and not a `part of a whole` of some structure. That's how I read it first. (yes, I know the word is the same, but you know what I mean.). Maybe something like `The overtone structure`

`If this happens a lot, we will notice chirping artefacts in the resynthesis` Can we hear an example of this?

`cross each other a points` typo

`Here are some partial tracks overlaid on the spectrogram, and the resynthesised sound:` no picture here?

`It didn’t do badly at all! Visually, we can see that things get a bit confused where things cross each other in the spectrogram, and that the algorithm struggles a bit with moving partials at higher frequencies (and note that we’re only getting stuff up to about 5kHz). However, the resynthesis is pretty convincing. Ok, something harder. Let’s do a field recording:` I try to avoid words like `things` and `stuff`, since they're easy to replace with what they're actually referring to and it makes it more clear.

`which shows that it’s finding it hard to model` I'm usually not opposed to personifications, but for this one, i might prefer something like `shows that the algorithm is designed to model sounds more like guitar--with clearer and more stable overtone structures.`

`and the depth has gone from the sound.` "depth" here is a metaphor so maybe there's a more objective way to describe what's lost? Also I am not hearing the sound, so maybe that will make it more clear.

`What we can do in the (frequent) case that sinusoidal modelling doesn’t capture everything interesting in our sound is to take the residual (the leftovers) and use this as as a separate layer` Rather than introducing the sines component as the "interesting" part of the sound and the residual as the other thing or after thought, i would just introduce the residual as another feature of the algorithm and let the user decide what is interesting.

`Alternatively we could do further decomposition, for instance, to try and separate transients from noise.` Yay! link to the flucoma tools that do that!

tedmoore (approved):

jamesb93:
Yep sorry both, I forgot to insert the audio before I added you for review...

jamesb93:
> `The partial structure should be pretty clear to see from the spectrogram.` I think this needs to be reworded so that it is more clear that by `partial` you mean an upper frequency component and not a `part of a whole` of some structure. That's how I read it first. (yes, I know the word is the same, but you know what I mean.). Maybe something like `The overtone structure`

`structure of partials` is what I went for

> `If this happens a lot, we will notice chirping artefacts in the resynthesis` Can we hear an example of this?

this text got culled (--:

> `cross each other a points` typo

solved

> `Here are some partial tracks overlaid on the spectrogram, and the resynthesised sound:` no picture here?

should be updated now

> `It didn’t do badly at all! Visually, we can see that things get a bit confused where things cross each other in the spectrogram, and that the algorithm struggles a bit with moving partials at higher frequencies (and note that we’re only getting stuff up to about 5kHz). However, the resynthesis is pretty convincing. Ok, something harder. Let’s do a field recording:` I try to avoid words like `things` and `stuff`, since they're easy to replace with what they're actually referring to and it makes it more clear.

I agree, more exact words good

> `which shows that it’s finding it hard to model` I'm usually not opposed to personifications, but for this one, i might prefer something like `shows that the algorithm is designed to model sounds more like guitar--with clearer and more stable overtone structures.`

_We can see that it models the dog barks using a large number of brief sinusoid tracks, which shows that it struggles to model that component, and won't be able to produce a compact representation of the signal._

Does that work better do you think?

> `What we can do in the (frequent) case that sinusoidal modelling doesn’t capture everything interesting in our sound is to take the residual (the leftovers) and use this as as a separate layer` Rather than introducing the sines component as the "interesting" part of the sound and the residual as the other thing or after thought, i would just introduce the residual as another feature of the algorithm and let the user decide what is interesting.

Yep: updated.

> `Alternatively we could do further decomposition, for instance, to try and separate transients from noise.` Yay! link to the flucoma tools that do that!

Internal links are now put in.

Thanks!

tedmoore:
> _We can see that it models the dog barks using a large number of brief sinusoid tracks, which shows that it struggles to model that component, and won't be able to produce a compact representation of the signal._
> 
> Does that work better do you think?

I think it is better. My concern is that saying it "struggles" suggests that the primary goal is an *accurate* resynthesis (and that there's a "not struggling" state to aspire to) which I think is not the primary goal for this tool since it will essentially never do a _very_ good job. This object is all artefact, so rather than guiding the reader to think that something is "succeeding" or "struggling" I'd rather try to frame it around whether the listener likes or dislikes what they're hearing. "Do I like it or do I not like it?" is creatively interesting where, "it struggles" sounds like "this tool is to be avoided in these situations".

It's like what I was saying above about the word "interesting" in regards to the sines + residual description.

Is this a useful sentiment to try to expound at this point in this document? I don't know, but that's my fully explanation of my thinking.

jamesb93:
> > _We can see that it models the dog barks using a large number of brief sinusoid tracks, which shows that it struggles to model that component, and won't be able to produce a compact representation of the signal._
> > Does that work better do you think?
> 
> I think it is better. My concern is that saying it "struggles" suggests that the primary goal is an _accurate_ resynthesis (and that there's a "not struggling" state to aspire to) which I think is not the primary goal for this tool since it will essentially never do a _very_ good job. This object is all artefact, so rather than guiding the reader to think that something is "succeeding" or "struggling" I'd rather try to frame it around whether the listener likes or dislikes what they're hearing. "Do I like it or do I not like it?" is creatively interesting where, "it struggles" sounds like "this tool is to be avoided in these situations".
> 
> It's like what I was saying above about the word "interesting" in regards to the sines + residual description.
> 
> Is this a useful sentiment to try to expound at this point in this document? I don't know, but that's my fully explanation of my thinking.

The primary _algorithmic_ goal is aimless or not situated in a musical question indeed. I think what needs to be communicated is the dog barking is a scenario in which sinusoidal modelling will produce artefacts and will fail because sinusoids are too simplistic to capture wiggly partials. That's the basic take home message in my eyes. My best guess woud be that this part of the text should be to point out the fact that the algorithm makes some assumptions about the sound, and they start to "crack" a bit under certain pressures. Then the question is left open for whether or not people like them.

PR #146, opened 2022-01-06
[Enhance] Auto-link headings

jamesb93:
Adds auto-link-headings to the build process to inject an <a> before each heading with a corresponding `id`.


jamesb93:
https://github.com/pngwn/MDsveX/issues/387

jamesb93:
These now work, but I need to find a good SVG to use for the link...

PR #148, opened 2022-01-07
[Content] AudioTransport Reference

jamesb93:
Where to look:`/reference/audiotransport/`

- add deps
- add css
- update mdsvex config
- update audiotransport reference


jamesb93:
@tedmoore this is ready to have a glance at from someone else if you have some time.

tedmoore (changes requested):
This is good. A few thoughts:

With the video of the peaks moving. I really wish there was sound! I know that I can audiate that spectrogram, but maybe not everyone can. Also, if you go back to it, I really like the idea of using the one where the three sines move different amounts and in different directions. 

I think it will help make more clear that this is happening on *magnitude* information in the frequency domain. (Maybe it'd be good to say that explicitly, I know you say spectrum, but I think it could be more explicit).

If it's possible to provide a little more intuition about *how* the algorithm works under the hood I know I'd be interested in that! I haven't read the paper, so maybe it's too mathy too quickly, but I think knowing that stuff is cool.

PR #150, opened 2022-01-10
[Content] Liminal Spaces

jdchart:
Liminal Spaces article as per requested.


jdchart:
Re videos - so I've just timestamped them for now, maybe better to get each separate on youtube but again, what i have is just ripped from the dialogues festival channel so dont know if that's great

jamesb93:
Ah it looks like its branched from Hayes, rather than from main?

PR #154, opened 2022-01-11
[Content] NoveltySlice Reference

jamesb93:
This adds a reference entry for the novelty slice object.

I've nabbed a lot of @weefuzzy's previous work and enriched it with some visualisations of my own confection. Would you be up for taking a squiz alongside @tremblap as this one is more knotty and it would be good to have both of your perspectives.


jamesb93:
> I love this, the example and all. it is still a bit cryptic and I wonder if the kernel size would benefit from another version of the graph with a 41x41 square drawn on it and moving along the line of difference, but maybe I get too used to sexy visualisations :D

I really want this too, but not sure it is cheap in terms of time for the payoff. Definitely doable but will take a few days to get it right perhaps.

PR #160, opened 2022-01-12
[Enhance] Made with FluCoMa Header Update

jamesb93:
Adds significant cosmetic updates to made with flucoma articles.

@jdchart do you think you could take a look at this and let me know what you think? You can check by having a peek at one of your articles that you wrote on the vercel preview.


jdchart:
Hey - flairs no longer a thing on mwf index ? Seems a shame.. 

I like the card border format thing with the artist bio - but Olivier's bio text seemes to have dissapeared ?

Like the green box for download also !

jamesb93:
> Hey - flairs no longer a thing on mwf index ? Seems a shame..

I forgot to reimplement it! I'll go back and do that.

> I like the card border format thing with the artist bio - but Olivier's bio text seemes to have dissapeared ?

Woops. I forgot to put it back in in the new format which uses a <slot /> to render some HTML rather than having the bio as a string variable.

I did this one later in the day so seems my marbles were gone a bit in the midst. Glad I put it to you for review @jdchart

jdchart:
The site's looking really nice though ! It's the small things

jamesb93:
@jdchart mind taking one look just to check I covered those errors?

In the future I'll do something a bit better for the artist profile, there is something not quite right that I cannot put my finger on.

jdchart:
All looks good to me.

For the artist profile, maybe it takes up too much space, like, seems weird to go onto an article and be met with half the screen covered by the top of Olivier's head..

Capture d’écran 2022-01-13 à 09 52 52

But honestly it looks good enough for now !

jdchart:
Capture d’écran 2022-01-13 à 09 52 59

even when scrolling you dont get the whole thing on the page.

But again, it doesnt look terrible. The border helps !

jamesb93:
> Capture d’écran 2022-01-13 à 09 52 59
> 
> even when scrolling you dont get the whole thing on the page.
> 
> But again, it doesnt look terrible. The border helps !

Hmm yeah...it works better for Hayes' because that picture has a more screen-like aspect ratio. I'll do some thinking and try something else maybe

jdchart:
To make things even more complicated, the upcoming articles feature two profiles ! Alice and Chris, Hans and Mark, Niamh and Alex ect

jamesb93:
That's good info - I'm whippin gup something now but need to go get vaccinated lol so I'll be back later to finish this work if you'd happy to review in the arvo?

jdchart:
No problemo!

jamesb93:
@jdchart i've made the component more real-estate friendly and smaller. It is also somewhat generalisable to a "duo profile" as we can slim them down and stick two next to each other. What do you think?

jamesb93:
I think we should also impose a character limit on the bios. Usually people have a smaller bio they can offer, like a 140 one and a 280 one...

jdchart:
I think this already looks a lot better

jdchart:
character limit, or like three lines displayed by default with expand option ?

jamesb93:
> character limit, or like three lines displayed by default with expand option ?

It could be expandable, thats not a bad idea at all! Only issue is that I'm wrapping up the entire container in an `<a>` (making it the whole box a hyperlink) which stops you from being able to select internal elements. Let's add this as an internal feature request and i'll loop around to it once I've got some other more pressing things out of the way.

Issue #91, opened 2021-12-02
Navigation and Discoverability


jamesb93:
I'm opening this issue as a sort of open question about how we can make the content more open to beginners who just want to browse as well as more directed for those who know what they want. One obvious thing to do is fuzzy searching. You type in "unsupervised" and all the algorithms that do unsupervised learning pop-up for example.

I'm thinking visual exploration too. We've discussed the possibility of some sort of decision tree that could be used to hone one's path through all the materials.

Another approach is something like this, not exactly, but like it. I don't like the fact that text is not all sitting flat.

https://bl.ocks.org/YungCrunchy/raw/a3ee962aefedef7eafa586646a80c29e/

jamesb93:
https://lunrjs.com/
https://fusejs.io/
https://www.npmjs.com/package/flexsearch
https://endler.dev/2019/tinysearch/

https://github.com/nextapps-de/flexsearch#nested-data-fields-complex-objects

Create a Corpus (js/py)
Make corpus searchable inside client-side

jamesb93:
https://svelte.dev/repl/a1b828d80de24f7e995b2365782c8d04?version=3.44.2

jamesb93:
https://github.com/mattyork/fuzzy

jamesb93:
https://www.npmjs.com/package/js-search

jamesb93:
https://awesomeopensource.com/project/ajitid/fzf-for-js

jamesb93:
https://github.com/nextapps-de/flexsearch#field-search is reall what we want I think... there is a section on tags below it too..

jamesb93:
Should investigate the possibility of adding all FluCoMa events and stuff to the database (introduced in #106) so that all the flucoma stuff is indexed.

jamesb93:
This is in a good place now, and instead of being a persistent issue the ongoing discussion should be moved to a repo wiki/discussion.

PR #163, opened 2022-01-13
[Enhance] Update reference page searchability

jamesb93:


jamesb93:
@weefuzzy @jdchart 

I'm trying to work towards my first major goal of the learn platform: [Navigation and Discoverability](https://github.com/flucoma/learn-website/issues/91).

Would you be able to take a look at `/references` and see if this interface is a bit more digestable than its previous iteration?

jdchart:
Format looks nicer than before (font + colours)
Much nicer that everything fits into one window, i can see everything at once.

Perhaps the blurb could appear in a popup when hovering on the name of the algorithm to give further explanation than just the algorithm name? But could clutter things.

jamesb93:
> Format looks nicer than before (font + colours) Much nicer that everything fits into one window, i can see everything at once.
> 
> Perhaps the blurb could appear in a popup when hovering on the name of the algorithm to give further explanation than just the algorithm name? But could clutter things.

Its a nice idea but I think it will clutter things. Also doesn't work for screen readers which I'm trying to support as best I can.

weefuzzy:
Agree it looks nicer. I miss what the search box could *potentially* have been (although I acknowledge that it wasn't that thing yet), but a vaguely intelligent way of filtering down that quite long list with some search terms would be really valuable. 

See the point about pop ups being difficult re clutter and screen readers, but I don't think this is insurmountable...

Anyway, in general I approve

jamesb93:
I was wary of duplicating the search up the top. Maybe I can mix elements of the previous and current iteration though?

weefuzzy (approved):
I think we can return to the question of snazzy search fields later, in the interests of manageable PRs n' all that. This is definitely a stylistic and organisational improvement, which was the point ❤️

jamesb93:
Cool, wangled it in for the impending PR storm.

PR #165, opened 2022-01-13
[Enhance] Add Search to references

jamesb93:
This adds a search bar to the references bar which uses fuzzy lookup against the title, category, tags and blurb. Seems to do the trick in not being too restrictive.

### DIFF
--- a/src/routes/reference/index.svelte
+++ b/src/routes/reference/index.svelte
@@ -1,10 +1,11 @@
 <script>
 	import { docs } from '$lib/app';
 	import ArrowRight from '$lib/components/ArrowRight.svelte';
+	import FuzzySearch from 'fuzzy-search';	
 
 	let references = docs.filter(d => d.section === 'reference');
 
-	const categorised = {
+	let categorised = {
 		'Analyse Data' : [],
 		'Analyse Sound' : [],
 		'Utility' : [],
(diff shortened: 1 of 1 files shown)


jdchart:
I was goign to say:
just brainstorming, when i search the page's size changes. Maybe instead of removing entries it could grey them out?

but then went back on to the page and think this is probably better

jamesb93:
I think greying is better actually. That way it's more about directing whoever is using it.

weefuzzy:
I'm ambivalent about greying. My concern is it would be less clear to look at

jamesb93:
Torn in two directions 😱

jamesb93:
@tedmoore @tremblap it would be good to hear your thoughts on this UX/UI issue.

The reference page has a search bar that lets you fuzzy filter the reference pages. It can do either two things:

1) Remove entries that are not captured by the search
2) Grey out entries that are not captured by the search

The current behaviour is 1, but it could be 2. What do you think?

https://learn-fjxra692z-flucoma.vercel.app/reference

tedmoore:
Right now the list of choices goes beyond the bottom of my browser window, so what I wouldn't want to need to do is to type in something and then scroll down to see what is *not* grayed out. I think if someone is typing in a search we should try to explicitly bring to them whatever it is they're looking for.

So my vote would be for removing the entries that don't match the fuzzy filter.

jamesb93:
> Right now the list of choices goes beyond the bottom of my browser window, so what I wouldn't want to need to do is to type in something and then scroll down to see what is _not_ grayed out. I think if someone is typing in a search we should try to explicitly bring to them whatever it is they're looking for.
> 
> So my vote would be for removing the entries that don't match the fuzzy filter.

Nice. Also maybe the reference blurb could be removed from the top, seeing as its already on the front page? that would make it easier to see everything at once.

tremblap (approved):
I'm in mixed feelings about this because I LOVE BOTH!

So let's not forget the THM: this is amazing either way.

1. the advantage of grey is that sometimes I'm not sure of what I search and the list is small
but
1. the advantage of pruning/removing is immidiate feedback

so it is a matter of tastes I reckon and if you guys have strong preferences I'll follow along with pleasure. our community might feedback (or not)

jamesb93:
I think most are in favour either fully or partially in having it just remove the entries opposed to greying them out. I think in my head grey seems nice but isn't explicit enough about what the search is doing/filtering. Before I merge I would like to adjust the top text, either by thinning it or cutting it out entirely.

PR #166, opened 2022-01-14
[Content] Fixed naming stupidness

jdchart:
the max project folder and project name were'nt the same, which caused an error when opening the project


jdchart:
Another bug which I thought i'd fixed, unable to load audio cause it consolidates into a different folder.... Working on it.

jdchart:
Ok, Max projects really annoying me now. 

surely the media in a project should be in the max search path and able to open up no!? Am I missing something ?
I thought i'd fixed this by just dropping the media into the patchers folder and not consolidating but it consolidates automatically on open.

Could @weefuzzy or @jamesb93 please check if this is also not working on your machine? for example downloading the pasquet patchers from this deploy, https://learn-git-content-maxproject-bug-fix-flucoma.vercel.app
and just trying to load one of the sounds in the first example?

weefuzzy:
> Could @weefuzzy or @jamesb93 please check if this is also not working on your machine? for example downloading the pasquet patchers from this deploy, https://learn-git-content-maxproject-bug-fix-flucoma.vercel.app
> and just trying to load one of the sounds in the first example?

Seems to work for me?

jdchart:
> 

Heffing hell. Updating Max to see if that's my problem...

jdchart:
Ok, I'm gonna guess I should have updated Max a long time ago...

weefuzzy:
Well, if it fails to work on older versions that *is* a problem, so we should probably check. I'll have a bash on 7 as well.

jdchart:
I'm going to guess it will work on 7, because i think i had quite an early version of 8...

jamesb93:
See how you go on 7 but if it doesn't work there we need to find a workaround.

weefuzzy:
It's related to how the child patcher is opened: if I open it from the project window, it uses the project search paths. When it's opened by pcontrol, it doesn't 🤦

jdchart:
Heffing heck. 

I mean, are we openly against distributing this in one folder rather than as a project? It's not as if there's thousands of files... @jamesb93 ?

jamesb93:
I'm not against having it in a folder. I reckon most people don't use projects. Will it work in a folder?

jdchart:
Great, it'll make it a lot simpler. It will work no problem!

I'll update then we can integrate.

PR #183, opened 2022-01-21
[Content] Grid Reference

jamesb93:
Adds reference entry for Grid﻿:

https://learn-git-fork-jamesb93-content-grid-reference-flucoma.vercel.app/reference/grid


tremblap (commented):
This is great, clear and nice with the widget. I would approve right away but I wonder if we should put a short paragraph (and a widget if easy) to explain the constraints in 1D and the oversampling (with a drawing of all the points that are left empty maybe) so we have all the elements explained there?  These latter 2 do not need interaction, maybe just a few examples as static images.

jamesb93:
There are technical barriers to achieving this right now. I couldn't (on my own) find a good way to leverage any existing libraries that perform linear assignment in js and weren't too heavy on the ol' CPU. So actually under the hood what happens is a simple interpolation of point-to-point. I could add some static images and loop back around if there were extra time.

jamesb93:
@tremblap I have re-requested review for you to check out the addendums of oversampling and extent if you have time at some point.

tedmoore:
This is super. A few feedbacks:

`This can be useful for compacting two-dimensional data into a uniform representation in which points are equally spaced from each other.`  I would use a word other than compacting. To me that implies (making more dense) and that's not always or really what's going on here. Maybe just "reorganising" (see the British English there! I'm learning!) (also I changed my computer's language settings so now I see red lines under American spellings!)

 `To do this, Grid analyses a dataset and determines a cost for moving (or assigning) each point two a position on a grid based on the euclidian distance between points.` "two" is a typo. Also, I would leave out the thing about "cost" since it doesn't get elaborated further. Many of our readers aren't going to be familiar with the type of algorithms that use "cost functions" so tossing "cost" in here is kind of a red herring.

`least worst` This is a little to cheecky and a little mind bending for my taste. I think just saying "best" is easier to read and clearer. or just: `Given these constraints, the algorithm will find the outcome in which the movement of points is minimised.`

`By default, the output grid is the number of dimensions` I think you mean number of points here. Also I think more accurate would be _the same as the number of points_ or _larger but as close as possible to the number of points_ in the input dataset, ensuring that Grid forms a square or rectangle.

tedmoore (approved):

PR #184, opened 2022-01-24
[Content] LabelSet Reference

jamesb93:
Adds a labelset reference

### DIFF
--- a/src/lib/components/DataSetView.svelte
+++ b/src/lib/components/DataSetView.svelte
@@ -5,31 +5,49 @@
 	}
 </script>
 
-<div class="table">
-	<div class="heading">Identifier</div>
-	<div class="heading">Data</div>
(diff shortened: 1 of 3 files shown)


tremblap (commented):
This is great. One more caveat we had to explain 903842390482 times: identifiers and labels are strings, even if they are made only of digits, so 60 is in effect "60" and very often "60" != 60

jamesb93:
yeah good catch. It is deffo important to explicate the types of the data under the hood.

tedmoore:
> identifiers and labels are strings

Is it more accurate to say that they're "symbols" in SC this is an important distinction and I think in Max we use `[fromsymbol]` to convert to an int (for example), so tying in to that terminology might be good.

tedmoore (approved):
see comment above

PR #187, opened 2022-01-24
Ampslice overview

tedmoore:
Is this material that we want to include somewhere? I tucked it under "Overviews" right now. I think looking at these envelope followers in this way would be nice.


jamesb93:
I would probably be convinced to merge this stuff with the reference rather than as an overview.

jamesb93:
Just as an FYI @tedmoore we need to have a protocol in place for committing and having images on the website. They need to be small, very small. More than 50-100kb is already too much.

I would be converting them to jpeg first and then compressing with something convenient like https://imageoptim.com/mac. This way the site runs faster, git doesn't get bogged down and it helps people view the content on slow connections.

jamesb93:
I will draw up a "contribting guide" so we can remind ourselves on some parameters for this sort of stuff collectively. The mp3s you made in the previous PR were good space savers, do you remember which settings you used @tedmoore ? I can bang one up for images then we'll have a central place to refer to.

jamesb93 (commented):
Needs discussion about where this material is most appropriate. I think it's discussing the _what_ of the algorithm, and would also support the current reference material. For those reasons I see this as a block that should be merged with the current reference.

tedmoore:
> The mp3s you made in the previous PR were good space savers, do you remember which settings you used @tedmoore ?

There were 128 kbps. Done in Reaper (which is [LAME](https://lame.sourceforge.io/)).

tedmoore:
@jamesb93, I integrated the ampslice "overview" with the pictures and stuff into the ampslice reference page. I commented out the widget (in lieu of the pictures) as discussed per the "slow" envelope not yet behaving properly. 

The text is mostly unchanged, just commented out or deleted text that was duplicated and interleaved a few things to try to make it smoother.

Happy to let this merge in if you feel good about it.

jamesb93:
> @jamesb93, I integrated the ampslice "overview" with the pictures and stuff into the ampslice reference page. I commented out the widget (in lieu of the pictures) as discussed per the "slow" envelope not yet behaving properly.
> 
> The text is mostly unchanged, just commented out or deleted text that was duplicated and interleaved a few things to try to make it smoother.
> 
> Happy to let this merge in if you feel good about it.

Yeah I feel good about it. Thanks for touching this up and making it more accurate and useful. I'll loop back around in docs v2 to do a widget but its just a time drain for now so it shall remain a bit shitty and commented out ;)

tremblap:
@jamesb93 james said:
> This way the site runs faster, git doesn't get bogged down and it helps people view the content on slow connections.

I cannot stress enough how much I agree with this.

PR #188, opened 2022-01-25
[Content] Loudness Reference

jamesb93:
- loudness skeleton
- loudness component skeleton
- nuance text


jamesb93:
@tremblap I'm asking you for review because you are the expert in loudness and have explained this several times to me. I'm struggling to find a way to leverage visual resources in this context so perhaps as part of your review you might suggest a straightforward scenario that I can widgetise that might help clarify the role of loudness for musicians. Easy in Max but a little hard in web-land to think of something convincing.

tremblap:
good draft. I've made a google doc with a revision and comments on the text.

https://docs.google.com/document/d/1x1lzapSG0Sv591gs8eU6XDGuIAkkeeVzRQvjW7vXbSg/edit?usp=sharing

Once edited, maybe sending it to Ted for more feedback

jamesb93:
> good draft. I've made a google doc with a revision and comments on the text.
> 
> https://docs.google.com/document/d/1x1lzapSG0Sv591gs8eU6XDGuIAkkeeVzRQvjW7vXbSg/edit?usp=sharing
> 
> Once edited, maybe sending it to Ted for more feedback

Thank you, all your changes are great.

@tedmoore would you mind taking a look at this and tell me what you think and feel? I'm not sure how to support this visually but I feel like it maybe needs it.

tedmoore:
I also think this is super. Very clear and appropriately concerned with the complexity of perception!

Visually, I can think of two graphics. One is of the Fletcher-Munson curve (or which ever curve is used for EBU standards). The second is a zoomed in picture of waveform showing the digital signal as a lollipop plot and the analog signal as a line. Where the line goes "higher" in amplitude (between samples) than the lollipop plot would suggest, showing why true peak can be important to calculate. It might even be the case that the analog line goes above an amplitude of 1, to make it "scary".

I also think it needs some explanation of what LUFS are. This could just be a link to an existing description, but it should be in there. Many readers won't know what that term means.

One suggestion. I wonder if this sentence: `For instant loudness, the recommended window size is 400ms updated every 100ms.` should be the second sentence of the paragraph? It directly explains `a small, configurable window of time`. Then in the next sentence you could be more explicit stating that in each of these time windows, Loudness measures...

In that same sentence, do you mean instance or instantaneous?

One typo: the word 'subjects'.

tedmoore (approved):
Left some comments above.

jamesb93:
> Left some comments above.

Thanks :) I'll crack on with your suggestions on the 'morrow

PR #192, opened 2022-01-27
[Content] Changed Alice's profile pic

jdchart:


jdchart:
If we could kindly merge asap :) @jamesb93

jamesb93:
In future if its a change like this feel free to merge it on your own as long as the build doesn't break on vercel.

PR #193, opened 2022-01-27
[Content] Copy flucoma.org made withs

jamesb93:
This takes the made with content from flucoma.org and puts it into the learn platform in an effort to centralise it.


jdchart:
Wow ! such fast. 

this looks good, although naturally it's kind of submerged with compositions now... Maybe we could have a "listen" section or something... but then a gain people may ask what's the difference between listen and made-with-flucoma?

jamesb93:
Without being dramatic, I think it pollutes/ruins the focus of the learn platform in this current structure. I am quite sensitive to this kind of thing 😄  IMO we either:

1. Roll these changes back and rename made with flucoma
2. Create an archive section?

We should relegate more detailed discussion to a formal meeting as per email thread.

tremblap (changes requested):
See email update - this is now probably obsolete

PR #196, opened 2022-01-28
NMFFilter reference

tedmoore:
Feedback requested.

One question I have: Would it be good to say something in the second sentence to the effect of, "It does this by doing non-negative matrix factorisation"? That (<--) sentence seems to not really be adding useful information, or at least leaves something to be desired. How does it do this...by which I mean what is the "matrix" if it's doing this one spectral frame at a time? Does the matrix it does this on have one of its dimensions = 1?

I looked at the C++ to see if it would glean any interesting way of thinking about it but that part of the code base is particularly nebulous to me.

The help file YAML says, "The FluidNMFFilter object decomposes and resynthesises an incoming audio signal against a set of spectral templates using an slimmed-down version of Nonnegative Matrix Factorisation (NMF)", but "slimmed-down" also seems vague enough to not be particularly useful.

Thoughts about that would be great. Maybe it's fine the way it is.

Thank you!


tremblap:
I love it

yes a second sentence on applying filters from nmf would be good. yes the process is trying to factorise the matrix which has only one frame long of activation. it is still valid math exercice to fit the 3 spectra proposed to the input spectra.

a few questions:
- the code is downloadable instead of being copy pastable like in other pages (with the Max/Sc/pd headers) I presume there is a reason
- i love the signal flow at the top but maybe we could make a colour thing to show that nmf is providing the bases - I didn't get that from the arrow
- every audio example has a link beside its header and that clutters the page - there must be a better way to have the header impact without all of them having links

jamesb93:
> * every audio example has a link beside its header and that clutters the page - there must be a better way to have the header impact without all of them having links

This is more for me than Ted to sort out. The link is superfluous you are right, and if their browser doesn't support the audio for some reason it will fall back to a link anyway.

> * the code is downloadable instead of being copy pastable like in other pages (with the Max/Sc/pd headers) I presume there is a reason

Just butting in where I'm not review requested but I think we should make consistent the style of multiple CCE downloads for code using the <Tab /> component. @tedmoore if you need an example see the installation pages. https://github.com/flucoma/learn-website/blob/9748a97005ecce7feee1668384b52ef6daf55d03/src/routes/installation/max.svx#L32 Ping me for any help you may need.

tedmoore:
> Just butting in where I'm not review requested
You were review requested!
Screenshot 2022-01-31 at 16 09 01

> we should make consistent the style of multiple CCE downloads for code using the component. @tedmoore if you need an example see the installation pages

Cool will do. This also answers @tremblap's query on why it was different.

> the process is trying to factorise the matrix which has only one frame long of activation. it is still valid math exercice to fit the 3 spectra proposed to the input spectra.

Excellent. I will incorporate.

tedmoore:
> * i love the signal flow at the top but maybe we could make a colour thing to show that nmf is providing the bases - I didn't get that from the arrow

@tremblap I'm having a hard time imagining what you mean here, can you mock it up for me?

src/routes/reference/nmffilter.svx
```diff
@@ -1,12 +1,124 @@
...
+In this example, we’ll use the bases from the analysis on the [BufNMF](/overviews/bufnmf) page. These bases contain spectral templates for the (1) kick drum, (2) snare drum, and (3) hi-hat components of the original drum loop.
+
+<Audio
+src="/audio/Nicol-LoopE-M.mp3"
+label="The original drum loop used to derive the bases from a BufNMF analysis"
+/>
+
+First, let’s send the original drum loop through NMFFilter in real-time using the bases derived from the drum loop analysis.
+
+### Audio stream output based on basis #1:
```
jamesb93, line 46: I personally am averse to using # in titles. I would say basis 1 (kick drum)

src/routes/reference/nmffilter.svx
```diff
@@ -1,12 +1,124 @@
...
+/>
+
+### Audio stream output based on basis #3:
+
+<Audio
+src="/reference/nmffilter/03_song_through_nmffilter_3.mp3"
+label="The output audio of the ensemble recording filtered by basis #3 of the drum loop BufNMF analysis"
+/>
+
+NMFFilter has relatively well separated the hi hat, kick drum, and snare drum sounds in this song (because those are the bases from the drum loop!). Of course there are other sounds in each of these audio signals as well, as all of the sound from the original file needs to be represented somewhere.
```
jamesb93, line 98: I think the paranthentical could be removed in favour of a comma.

`in this song, because those are the bases from the drum loop`

tremblap:
> > * i love the signal flow at the top but maybe we could make a colour thing to show that nmf is providing the bases - I didn't get that from the arrow
> 
> @tremblap I'm having a hard time imagining what you mean here, can you mock it up for me?
Screenshot 2022-02-02 at 13 31 33

tremblap:
that way we see that the imported bases (in number and quality, the 3 rect could be 3 filter curves and colour coded) are important in quality and numbers

tedmoore:
> that way we see that the imported bases (in number and quality, the 3 rect could be 3 filter curves and colour coded) are important in quality and numbers

I like this very much!

tedmoore:
This said it failed but all i did was change one picture that worked for me running `pnpm run dev` and `pnpm run build`, so I merged.

jamesb93:
> This said it failed but all i did was change one picture that worked for me running `pnpm run dev` and `pnpm run build`, so I merged.

When vercel or github actions builds it does it against the tip of svelte and sveltekit which just introduced a breaking change to the API. I've fixed it now so shouldn't happen again.

PR #197, opened 2022-01-31
added tabs for downloading example code

tedmoore:
@tremblap Are there other places where code is copy-and-paste-able? I found some tabs on the website but they just hide and show links.

Is this what you had in mind?

### DIFF
--- a/src/routes/reference/nmffilter.svx
+++ b/src/routes/reference/nmffilter.svx
@@ -16,6 +16,7 @@ category: Transform
     import IndentNote from '$lib/components/IndentNote.svelte';
     import Image from '$lib/components/Image.svelte';
     import Audio from '$lib/components/Audio.svelte';
+    import { Tabs, TabList, TabPanel, Tab } from '$lib/components/tabs/tabs';
 </script>
 
 <IndentNote type='pointer'>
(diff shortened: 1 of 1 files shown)


jamesb93:
> 

The tabs can contain anything, whatever is nested in the tab is rendered when that tab is active. If you need 5 minutes of walkthrough on how it works let me know.

tedmoore:
@jamesb93 I implemented the tabs in this branch. I just wondering if this is what @tremblap was suggesting when he said:

> the code is downloadable instead of being copy pastable like in other pages (with the Max/Sc/pd headers) I presume there is a reason

in https://github.com/flucoma/learn-website/pull/196

There's basically just one sentence and link per tab, so I think it ends up taking up more room as a tabbed thing than just a couple links.

If the idea is to have the code be copy-and-paste-able, would that be for SC only (and one would download the `.maxpat`) or do we want to put the copy-able hex (?) code thing for max patches in there as well? I didn't see that elsewhere on the website right now, so just trying to feel out what the house style might be to try to be consistent going forward.

jamesb93:
Is there anything you needed me to look at further here?

tedmoore:
nope. thanks!

PR #204, opened 2022-02-04
[Content] BufScale reference

jamesb93:
Adds the BufScale reference.﻿


jamesb93:
@tedmoore this is quite a lean reference. BufScale is quite narrow in what it does and I think perhaps to an extent self explanatory. If you can think of some more interesting ways to invite people into the content let me know. Otherwise if you approve go for squash merge :)

tedmoore (changes requested):
Yes, I think this is appropriately short. It will be nice that the JSON will render the arguments into the HTML as well.

`The BufScale object is a tool for scaling the value of data stored in a buffer.` I think this is a bit misleading as it sounds like it does the operation in place, when in fact, there is a source buffer and a destination buffer.

It might also be good to highlight that the operation can be done on a sub section of the buffer (and I'm guessing that _that_ is the section of the buffer that actually ends up in the destination buffer?).

I know it will be covered in the arguments list, but it also would be good to have just one sentence that **clipping** is an option.

jamesb93:
> Yes, I think this is appropriately short. It will be nice that the JSON will render the arguments into the HTML as well.
> 
> `The BufScale object is a tool for scaling the value of data stored in a buffer.` I think this is a bit misleading as it sounds like it does the operation in place, when in fact, there is a source buffer and a destination buffer.
> 
> It might also be good to highlight that the operation can be done on a sub section of the buffer (and I'm guessing that _that_ is the section of the buffer that actually ends up in the destination buffer?).
> 
> I know it will be covered in the arguments list, but it also would be good to have just one sentence that **clipping** is an option.

All good suggestions will work on these.

PR #207, opened 2022-02-07
[Feature] Generic {Multi}Slider component

jamesb93:
As requested here is a slider component. I had thoughts about building a multislider, but it seems easy enough and flexible to just compose series of sliders in whichever one sees fit. They all bind themselves to some bit of data such as an array element or a single value so its quite easy to map them where one wants to. Might demand some more explaining but maybe they are visually enough for you @tedmoore.

A demo should pop up here:

https://learn-git-fork-jamesb93-feature-multislider-flucoma.vercel.app/demos/multislider

with code conveniently shown below:

```sv
<script>
    import * as _ from 'lodash';
    import * as d3 from 'd3';
    import { Slider } from '$lib/components/multislider/multislider';

    const accent = d3.scaleOrdinal(d3.schemeAccent);

    let value = 1;
    let values = new Array(10).fill(0).map(x => Math.random())
</script>

<h2>A single slider</h2>
{ value }

<Slider 
bind:value={value}
min={0}
max={9}
steps={10}
/>

<h2>Composing multiple sliders together</h2>

<div class="row">
    {#each values as x, i}
    <div class="col">
        {x.toFixed(1)}
        <Slider 
        bind:value={values[i]}
        min={0}
        max={6}
        steps={10000}
        height={'100%'}
        color={accent(i)}
        />
    </div>
    {/each}
</div>

<style>
    .row {
        display: flex;
        flex-direction: row;
        gap: 0.5em;
        height: 200px;
    }

    .col {
        display: grid;
        grid-template-rows: max-content auto;
        place-items: center;
    }
</style>
```


jamesb93:
Actually making this draft, I realised I didn't figure out bipolarity!

tedmoore:
@jamesb93 this may no longer be necessary if i'm going to try with p5 first!

jamesb93:
Oh yeah! Okay I'll keep it as draft and we can reopen one day if needed. Do let me know how you go I'm intersted.

jamesb93:
I'll come up with something superior to this.

PR #210, opened 2022-02-09
[Enhance] Related articles under table of contents

jamesb93:
This adds a "related" section under the table of contents. It basically pulls any internal links out of the document at build time and then creates a small json bundle that can be accessed client side. 

My questions are, is this useful? How can we make it even more useful? I imagine `seealso` can become a part of this when we eventually get the DOCS producing raw JSON.


jdchart:
This is excellent

Issue #201, opened 2022-02-02
new entry 


tremblap:
a single entry in learn, maybe in articles, where we talk about where is the long list of made-with, so if someone searches for made-with they are pointed at the right place. that should go in the 3rd tab (stories, projects, whatever we call it)

jdchart:
Right, just pointing to the www or also with a list (i believe we wanted to avoid duplicates)?

jamesb93:
Can you clarify what this suggestion is? I'm finding it a little hard to understand @tremblap

Issue #200, opened 2022-02-02
top of learn.flucoma.org


tremblap:
we talked about the 3 elements and their relative importance

logo (always there, on the left)

3 items (last resource is to hamburger them)

search (always on the right, shrinks as much as possible)

tremblap:
as a thinking space:
http://www.wekinator.org/

jdchart:
???

is the wekinator website here as an example of what NOT to do ?

jamesb93:
I think a search bar should not shrink too much, but I agree on everything else in terms of preferencing behaviour of the UI. On a similar note I think that perhaps the search bar could organise the results for you like this:

https://www.manomano.co.uk/

to help make the experience of navigating through the results tighter.

It should also have a search icon _before_ the input text to help denote that its a search bar and you are using it to search. I have a dream of a UI where it expands when you focus on the field input and collapses when youre not. I'd have to see how it looks on mobile.

Issue #199, opened 2022-02-02
headers for the learn platform


tremblap:
We discussed, a thread was started, and we need to decide.

 We change the names of the 3 things. Now, this is where we need to agree as it is more controversial:

  1) Learn/ reference / explore

  2) Overviews / References / Projects

  Contenders for item 3: podcasts/stories

tremblap:
As I said in the thread, I love stories but it feels hippie and a bit corporate-coopting strangely… but I could be swayed

jdchart:
I still vote: learn/reference/explore

overviews is too vague for me. projects not good cause not all the podcasts are projects. stories no.

tedmoore:
For the first one, I do like `learn`

For the second one, it seems `reference` is pretty well liked, including by me

For the third one, I actually still like `projects` or maybe `examples`. I'm not bothered by having podcasts fall under either `examples` or `projects`

jamesb93:
my vote is for learn / explore / reference. In that order as well, such that learn and explore are coupled together whenever we mention it and reference being on the end. This could be supported by some clever UI updates and a bit of guard railing for a user entering the platform. I think that this might suggest a better way of approaching all the content that suggests three unique streams, with their own ways of maneouvering around the content.

jamesb93:
Yes you read me right.

tremblap:
this looks better - it would be great that the logo stayed there when the hamburger arrives though

tremblap:
should I put another ticket?

jamesb93:
> this looks better - it would be great that the logo stayed there when the hamburger arrives though

I'm not sure what "this" is. AFAIK nothing has been changed in regards to the top navigation.



> should I put another ticket?

Already a ticket here: https://github.com/flucoma/learn-website/issues/200

Issue #190, opened 2022-01-27
Footer


jdchart:
Just noticed at the bottom of pages there's this footer:

Last modified: Thu Jan 27 09 by Jacob
Commit: 982af2
Edit File on GitHub

Is this intentional ? Date and author i can understand but do people really need to know what commit it is...?

jamesb93:
Not necessarily. I see that pattern a lot for documentation type stuff but I potentially agree with you that it might even put people off?

https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics (for example).

jdchart:
fair enough. I dont think it would go so far as to put poepl off, i just personnaly found it a bit weird. i guess i never read the footers of websites!

jamesb93:
Less content on the page means less to render, and less to read. Let's remove it and see if anyone cares.

PR #213, opened 2022-02-11
[Enhancement] Various Learn Enhancements

jdchart:
This is going to be the PR that implements the general changes discussed in these issues: 

#199 
#200
#201 
#190 

Not ready to merge yet but will be soon.


jdchart:
First, #201 , @tremblap I have simply added a link at the top of made-with directing towards a full catalogue of pieces made with the tools: https://learn-git-front-page-flucoma.vercel.app/madewithflucoma 

is this sufficient or do you want an entry under made-with?

And i suppose we must therefore remove the articles concerning the 2 gigs, even if i think it's a shame. I can link in the same manner, or again, make an article that directs people to all the different types of content on the www.

jdchart:
Started working on a new homepage layout which i think will be better at showing what the site has to offer - although the implementation of it is a complete shit-show. Just prototyping ideas.

jamesb93:
I would appreciate no one elses merges this till I have taken a look. Changing the names is more complicated than changing them on the surface of the HTML.

Also https://github.com/flucoma/learn-website/issues/190 is redundant now.

jamesb93:
I've opened a new PR for #199 here https://github.com/flucoma/learn-website/pull/218

jdchart:
Yep, will ping when need input, and good call to move #199 to #218 @jamesb93 . 

Indeed I was wanting to look at the front page stuff with you James, this was just to get an idea. Notably accessing the global layout css - I'd perhaps like things to be wider, less margin. And we can discuss how we would want to feature things - at the moment its just random on each page load. And perhaps talk about being able to define a featured image for each page so that there's potentially more eye-catching things on the front page. But i do like this layout where we can see more of everything at the same time (without it being chaotic).

jdchart:
@tremblap for #201 , let me be more clear:

for the moment, I've just added a line of text and a link at the top of the made with flucoma page that points towards the www. learn catalogue of pieces.

is this sufficient, or do you want a full 'article' in the made with flucoma section that deals with this, and perhaps also other things that are on www like the gigs?

Also, there are two full entries in made with that deal with the gigs, but of course there's overlap with what's on the www website (even if I find the learn entries to be more complete). For the sake of not havign duplicates do you want me to remove these entries?

jamesb93:
> Yep, will ping when need input, and good call to move #199 to #218 @jamesb93 .
> 
> Indeed I was wanting to look at the front page stuff with you James, this was just to get an idea. Notably accessing the global layout css - I'd perhaps like things to be wider, less margin. And we can discuss how we would want to feature things - at the moment its just random on each page load. And perhaps talk about being able to define a featured image for each page so that there's potentially more eye-catching things on the front page. But i do like this layout where we can see more of everything at the same time (without it being chaotic).

I agree with all of this :) Wider, pictures that get gobbled up as a preview. Let's talk about it more when you have some brainspace.

PR #221, opened 2022-02-16
[Learn] Seeding Activations and Bases

tedmoore:


src/routes/learn/seeding-nmf.svx
```diff
@@ -7,7 +7,175 @@ tags:
...
+    // import ResourceLink from '$lib/components/ResourceLink.svelte';
+</script>
+
+<IndentNote type="pointer">
+
+This overview builds on information found in the [BufNMF Overview](/learn/bufnmf). Please visit that page first to learn more.
+
+</IndentNote>
+
+Non-negative matrix factorisation can be "seeded," meaning that it is given some information to begin the process from. You can provide bases and/or activations that you think will "point the algorithm" in the direction toward what you're hoping it will decompose. Without seeding, NMF activations and bases are both initialised with random values.
```
jamesb93, line 32: I would remind them of the process, even a key word like "decomposition" might be helpful here

jamesb93 (changes requested):
The graphics could be made slightly easier to read by making any lines thicker. I'd twice as thick if it doesn't obscure them but anywhere between what they are now and double would be great. Otherwise its all very very clear.

tremblap:
too late for review it seems but this is fantastic! It gets a bit jargony but hey, it is quite something. I think that the oversampling bundling of the guitar is still missing as one last example though, as a form of synthesis of machine listening to sieve.

tremblap:
also there is a white lie in there - as it is merged I'll do a bug report

PR #222, opened 2022-02-17
Layout Infrastructure

jdchart:


jdchart:
To do:
- wrap the reference and installation indexes with he css that they used to inherit.
- currently there's a general "Content" layout - find a way of allowing the different types (reference, learn, explore) to apply their small css changes (the best way, I think, would be to have a component for each type that is imported on each page, and run at the end, and tat would change the css for the various div classes as needed...).

jamesb93:
Just as a heads up, the convention for layouts is that they are lower case :)

PR #226, opened 2022-02-17
[Content] Sines Reference

jamesb93:
Creates a reference for the sines algorithm.

@tedmoore, mind having a look?﻿


tedmoore (changes requested):
Looks great. A few thoughts.

* I notice that the name of the page often links to itself, such as the first word of this article. I find this confusing--it looks like a link and therefore seems like it should link somewhere, so when it doesn't it seems like a broken link. Is there a philosophy around this?
* `Anything that can’t be easily resynthesised in this way is separated and classes as a “residual” component.` I would delete `and classes`
* `Don’t get too excited though, this is the perfect scenario for [Sines](https://learn-git-fork-jamesb93-content-sines-flucoma.vercel.app/reference/sines/), because the sinusoid is highly differentiable from the noise and isn’t made more complicated by things modulating through time.` I think the last clause of this sentence should be it's own sentence. I think it's a sufficiently different idea from the first part of the sentence.
* `cymbals and hi-hat` I'm pretty sure there's just the hi-hat in there.
* `For more detailed information on how this method of decomposition works, take a look at the [sinusoidal modelling](https://learn-git-fork-jamesb93-content-sines-flucoma.vercel.app/learn/sinusoidal-modelling) article.` These two pages seem very similar, this one essentially being a shorter version of the other one with a different example. It makes me wonder why there are the two different pages?

jamesb93:
> * I notice that the name of the page often links to itself, such as the first word of this article. I find this confusing--it looks like a link and therefore seems like it should link somewhere, so when it doesn't it seems like a broken link. Is there a philosophy around this?

A weak one. If you link to the page you are already navigated to then navigation doesn't happen. We could just operate on the principle that you don't link to yourself going forward?

> * `Anything that can’t be easily resynthesised in this way is separated and classes as a “residual” component.` I would delete `and classes`

agree

> * `Don’t get too excited though, this is the perfect scenario for [Sines](https://learn-git-fork-jamesb93-content-sines-flucoma.vercel.app/reference/sines/), because the sinusoid is highly differentiable from the noise and isn’t made more complicated by things modulating through time.` I think the last clause of this sentence should be it's own sentence. I think it's a sufficiently different idea from the first part of the sentence.

okidoki

> * `cymbals and hi-hat` I'm pretty sure there's just the hi-hat in there.

I'll have a listen back

> * `For more detailed information on how this method of decomposition works, take a look at the [sinusoidal modelling](https://learn-git-fork-jamesb93-content-sines-flucoma.vercel.app/learn/sinusoidal-modelling) article.` These two pages seem very similar, this one essentially being a shorter version of the other one with a different example. It makes me wonder why there are the two different pages?

We discussed this in sync (01/03/2022) and agreed there is some overlap to address. I agree, this could be merged with the reference.

tedmoore:
> > * I notice that the name of the page often links to itself, such as the first word of this article. I find this confusing--it looks like a link and therefore seems like it should link somewhere, so when it doesn't it seems like a broken link. Is there a philosophy around this?
> 
> A weak one. If you link to the page you are already navigated to then navigation doesn't happen. We could just operate on the principle that you don't link to yourself going forward?

I see. I think it's confusing that we have links that go nowhere, so yeah, going forward we can just not link to the page things are on. and if we see this in any documents as we are working on them might as well remove them!

tedmoore:
fine with me.

jamesb93:
I'll just leave this one as draft and merge the other one in then.

PR #231, opened 2022-02-24
[Layout] Front Page

jdchart:


jdchart:
not to be merged yet of course @jamesb93 , but just to show you how things are progressing. for the moment a new metadata file is created that will be used ny the front page (that includes featured media - if no featured media is given, a random one is chosen)

jdchart:
As @jamesb93 explained in #238 that PR will be merged into this one.

@tedmoore @weefuzzy @tremblap would love to get your thoughts on this front page design. 

At the moment, the featured stuff is easily manually set in a static json file (allowing people to change the featured content in the future).

weefuzzy:
I assume some of this would change due to @jamesb93 work on a menu etc? 

I like the 'Featured' block. I like the for the pieces in the Learn and Explore blocks, but I'm not sure how useful the blocks themselves are, because (a) we have to scroll and (b) it's not then clear if we're seeing curated highlights or the whole of a category. Perhaps better to focus on having quick ways to get to the index pages for the categories and coming up with little images to accompany the index entries. 

Likewise for the 'algorithms' block, but it has no images, so I like it even less

jdchart:
> I assume some of this would change due to @jamesb93 work on a menu etc?

Yup, the only part of this to pay attention to is the content on the front page.

> I like the 'Featured' block. I like the for the pieces in the Learn and Explore blocks, but I'm not sure how useful the blocks themselves are, because (a) we have to scroll and (b) it's not then clear if we're seeing curated highlights or the whole of a category. Perhaps better to focus on having quick ways to get to the index pages for the categories and coming up with little images to accompany the index entries.

I do get what you mean by it being unclear if this is curated highlights of the whole of a category. However - I think it's better to have on the front page _content_ rather than _categories of content_. I think it looks much more engaging and lively. I think the issue is communicating _why_ that content is there... And it's true that as it stands, there is essentially no reason... Perhaps "previously featured _category_" ?

I'm also have no inherent aversion to scrolling...

Yes, I think the next PR will also be about making index pages look a bit more like this (with pretty pictures and stuff).

jdchart:
@weefuzzy would this titling (or something in this spirit) resolve some things or are you still not a fan? 

@jamesb93 also suggested perhaps having the short section description next to the "previously featured" headings also.

weefuzzy:
Well, I guess I'd like them to be much smaller if it's a 'previously <x>' kind of thing. I'm not allergic to the idea of a 'algorithm of the day' sort of thing either, but I don't think it will be found easily all the way down there. I do think that, for a front page, you want to be able to communicate what's on offer clearly within the available view port *or* to make it really obvious that there will be rewards for scrolling.

jamesb93 (approved):

PR #235, opened 2022-03-02
[Content] OnsetSlice Reference

jamesb93:
Adds a reference entry for OnsetSlice.

@tremblap do you mind taking a look?


jamesb93:
> * a simple sentence on the fact that most metric will be considering upwards changes in the spectrum

What do you mean here? I think by virtue of the fact that the algorithm won't detect a new slice until the previous frame has gone under the threshold means that you only get upwards changes which is not tied to the metric. Or I don't know something about the metrics and how they work which is likely. 🤷 

> * a simple sentence per metric to say how it 'models' the difference so people don't have to go to the actual paper if they are afraid of maths

Are you able to help me draw up a musicianly interpretation of each metric? I think some of them are quite cryptic still. I can also refer to the paper to see if itll help me in that regard. I agree though that some outlining of the various metrics would be good.

> * a reference to the paper that has all of the metrics (I thought that we had one that had all of them as we plunder the SC version of onset and they had a paper)

That one should be in the related resources at the bottom. DAfX paper archive is offline and might have been for some time so I couldn't trace which one was referenced in the docs. From the name of the metrics this seems to be the one though.

PR #236, opened 2022-03-02
[Enhance] Waveform Views for Audio Component

jamesb93:
Adds waveforms to all instances of <Audio /> component.

@tedmoore and @jdchart what do you think? A good example to peek at is the HPSS reference.﻿


jdchart:
Think it looks nice ! Looks nice in situ of an explore article too, cf Pasquet

jdchart:
jacob hart approves this

PR #238, opened 2022-03-02
[Enhance] Navigation and Header Layout

jamesb93:
Hi all who are tagged (@jdchart, @tedmoore, @tremblap , @weefuzzy)

I've been working towards not having such a cluttere design, and following some more modern design principles.

This will integrate with Jacobs PR (#231) eventually so that is worth checking out.

Main points are:

1. Header is smaller to get out the way
2. Navigation is pushed to the side hamburger menu (meaning the descriptive text for each section is contained to the navigation rather than the front page)
3. Search is slightly changed (though that will get changed up later too I imagine).

Keen to hear your gut reacitons to it.


jdchart:
Looks really great! A few things you're probably already aware of for ironing:

- On my screen, it leaves the screen :'(
- The cursor is a hand pointer thing when off the menu which makes me sad
- Menu doesn't yeet when page changes (but you know that)
- not currently stopping scroll on the main page

It's going to take some getting used to there not being the links at the top, but that's cause ive been going on this website everyday for the past couple of months. I think this is definately a step in the right direction.

Capture d’écran 2022-03-03 à 10 07 47

jamesb93:
> Looks really great! A few things you're probably already aware of for ironing:
> 
> * On my screen, it leaves the screen :'(

What do you mean?
> * The cursor is a hand pointer thing when off the menu which makes me sad
I prefer this, it shows that that area is clickable and will be helpful to users. Seen that pattern elsewhere too.

> * Menu doesn't yeet when page changes (but you know that)
Yep, impending fix on the way!
> * not currently stopping scroll on the main page
Weird! seems like I need to be more dilligent here. Thanks for flagging it.

> It's going to take some getting used to there not being the links at the top, but that's cause ive been going on this website everyday for the past couple of months. I think this is definately a step in the right direction.
I agree, feels "empty".

> 
> Capture d’écran 2022-03-03 à 10 07 47

jdchart:
> What do you mean?
> 
> >

Look at the screen shot, i dont get all of the blurb for explore

jamesb93:
> > What do you mean?
> > >
> 
> Look at the screen shot, i dont get all of the blurb for explore

You look at the screen shot

jdchart:
Looks very noice.

Still a small issue of description exceeding window, see screenshot:

Capture d’écran 2022-03-04 à 08 38 28

weefuzzy:
Looking nice and clean 

1. The site should say *what* it is, at least! If one is deposited onto a inner page via URL, we need to make sure it's easy to work out what to do from there 
2. Now that the burger is a Very Important Thing, I wonder if it should be more prominent by being (a) bigger (b) not in the little header (c) a position:fixed thing that follows down the page
3. The horizontal position of the burger seems weird, especially as I resize the screen

jamesb93:
> 1. The site should say _what_ it is, at least! 

So keep some sort of "FluCoMa" in the top bar? Can do!

> If one is deposited onto a inner page via URL, we need to make sure it's easy to work out what to do from there

Perhaps that will be solved by having the burger by extra prominent? See comments below.

> 2. Now that the burger is a Very Important Thing, I wonder if it should be more prominent by being (a) bigger (b) not in the little header (c) a position:fixed thing that follows down the page

Yes so perhaps it actually floats over everything, and when the navigation is expanded it sits on top?

> 3. The horizontal position of the burger seems weird, especially as I resize the screen

Solved by a fixed position.

jamesb93:
> Looks very noice.
> 
> Still a small issue of description exceeding window, see screenshot:
> 
> Capture d’écran 2022-03-04 à 08 38 28

I see all of it on my 13" but perhaps yours is non retina which may be the issue?

weefuzzy:
> So keep some sort of "FluCoMa" in the top bar? Can do!
Some mention of learning too 😁 

> Yes so perhaps it actually floats over everything, and when the navigation is expanded it sits on top?

Sounds good – although I guess it would need to be different on very narrow displays like phones

jdchart:
> I see all of it on my 13" but perhaps yours is non retina which may be the issue?

I have 13,3 inch (1280 x 800) - macbook pro mid-2012, so I'm assuming no retina.

weefuzzy:
We should definitely check on Windows too 😬 

I think a simple part of the solution is to make those descriptions more pithy

jamesb93:
> We should definitely check on Windows too 😬
> 
> I think a simple part of the solution is to make those descriptions more pithy

We could take the pithy ones from Jacob's PR. "Explore doing a flucoma" that kind of thing

tremblap:
I know I'm late to the party, but I really love the site now. Maybe the 3 headers at the top, above <Featured> could be a little bigger so they are a bit more assertive, otherwise it is perfect.

PR #239, opened 2022-03-03
[Content] MelBands Reference

tedmoore:
@weefuzzy, I wanted to ask your review on this one as well since we had a chat about it all. Thanks!


src/routes/reference/melbands.svx
```diff
@@ -1,8 +1,86 @@
 ---
 title: MelBands
-blurb: Amplitude for a number of equally spread perceptual bands
+blurb: Magnitudes for a number of perceptually-evenly spaced spectral bands
```
weefuzzy, line 3: Personally I'd hedge 'perceptually-evenly spaced' because the mel scale is an approximation, and possibly not even an especially good one. Expectation management n' that

src/routes/reference/melbands.svx
```diff
@@ -1,8 +1,86 @@
...
\ No newline at end of file
+---
+
+<script>
+    import Image from '$lib/components/Image.svelte';
+    import YouTube from '$lib/components/YouTube.svelte';
+    import { Tabs, TabList, TabPanel, Tab } from '$lib/components/tabs/tabs';
+</script>
+
+`MelBands` returns a Mel-Frequency Spectrum which is a [histogram](https://en.wikipedia.org/wiki/Histogram) of FFT bin magnitudes bundled according to their relationship to the [Mel scale](https://en.wikipedia.org/wiki/Mel_scale). The Mel scale represents frequency space logarithmically, mimicking how humans perceive distance in pitch space. (The name "Mel" derives from the word "melody".)
```
weefuzzy, line 20: Don't know that 'histogram' is especially enlightening as an explanation here. It's not a conventional histogram like people learned at school because of non-rectangular, overlapping bands. Also, the DFT magnitude spectrum is a 'histogram' to exactly the same extent, so it doesn't really distinguish the mel-bands case. 

Possibly worth noting that there are other perceptually motivated scales, deriving from different experiments / theories of hearing, such as Bark bands or Equivalent Rectangular Bandwidth (ERB). "Human and Machine Hearing" by DIck Lyon has a good survey and critique of a bunch of these.

src/routes/reference/melbands.svx
```diff
@@ -1,8 +1,86 @@
...
+
+<YouTube
+url='UCLlVAj0PPY'
+/>
+
```
weefuzzy, line 25: Nice animation, perhaps explain what's happening in a very brief alt-texty way too

src/routes/reference/melbands.svx
```diff
@@ -1,8 +1,86 @@
...
+One can choose whether to `normalize` triangle filters used to compute the MelBands. By default (unnormalized, `normalize` = 0), the triangle filters account for how many FFT magnitudes they sum together by dividing the sum by the number of contributing FFT magnitudes, essentially averaging their contributions. When the triangle filters _are_ normalized (`normalize` = 1), averaging does not occur and the magnitudes of the higher MelBands end up being larger values (disproportionately so) becuase they're summing together more FFT magnitudes. In the charts below, notice how different the unnormalized and normalized triangle filters are and when normalized, how much larger the magnitudes in the higher MelBands become.
```
weefuzzy, line 28: normalization is *on* by default

weefuzzy: I think the explanation is backwards too. When normalization is on, the bands are scaled proportional to the linear area they occupy, so the higher bands are turned down

src/routes/reference/melbands.svx
```diff
@@ -1,8 +1,86 @@
...
+
+### Unnormalized Triangle Filters (the default)
+<Image
```
weefuzzy, line 31: perhaps the figures would benefit from having a dB amplitude scale?

tremblap:
great widgets and explanation - I propose a few changes and reorganisation below but otherwise this is great

represents frequency space logarithmically -> not strictly true, that would be midicents. it is lin at the bottom and exp from 500 ish and more than log later up, which mimicks the perception of pitch on humans...

(The name “Mel” derives from the word “melody”.) - fun trivia but I'd remove - or put earlier. I am allergic to end of paragraph parenthesis stuff. if it is worth saying, let's say it for realz

'Because the Mel-Frequency Spectrum relates to how humans perceive the spectral shape of sound, it can make for a more relevant audio descriptor (than FFT spectra) for considering how humans compare and hear different sounds.' this is key and I'd put at the end of the first paragraph, before we get mathy/fft

tedmoore:
> it is lin at the bottom and exp from 500 ish and more than log later up, which mimicks the perception of pitch on humans...

@tremblap I think that ours is actually not linear at the bottom. The librosa page says that by default they do have it be linear up to 1k and log after that, but they also have an alternative option (based on the Hidden Markov Toolkit) that is just log (which is different from ours, but of the same flavour).

https://librosa.org/doc/0.9.1/generated/librosa.mel_frequencies.html?highlight=librosa%20mel_frequencies

If ours was truly linear up to a point and log after that it would need to be a piecewise function, but it is not. 

https://github.com/flucoma/flucoma-core/blob/cdbe03c292dbf220dbc46c3c41aedc60451fe9cb/include/algorithms/public/MelBands.hpp#L36

Perhaps there's something i'm missing?

PR #240, opened 2022-03-04
Pitch Reference

tedmoore:


src/routes/reference/pitch.svx
```diff
@@ -8,4 +8,44 @@ flair: reference
...
 ---
 
+<script>
+  import ResourceLink from '$lib/components/ResourceLink.svelte';
+  import Image from '$lib/components/Image.svelte';
+  import Audio from '$lib/components/Audio.svelte';
+</script>
 
+
+FluCoMa's Pitch object returns both pitch and confidence values. When no pitch can be detected, a pitch of 0 Hz is returned (or -999.0 when the unit is in MIDI note mode).
```
jamesb93, line 18: Would prefer something agnostic like The Pitch algorithm in the FluCoMa toolkit

src/routes/reference/pitch.svx
```diff
@@ -8,4 +8,44 @@ flair: reference
...
+
+The confidence output is a value between 0 and 1 indicating how confident the algorithm is _that there is_ a pitch. In effect this can be an estimation of how "noisy" (closer to 0) or "harmonic" (closer to 1) the spectrum is.
```
jamesb93, line 20: The confidence output is a value between 0.0 and 1.0 and indicates how confident the pitch algorithm is in its estimation. It is a useful piece of information that can help us disregard noisy moments in time (closer to 0), where pitch would be hard to discern, as well as giving us a general indication of how harmonic (closer to 1) the spectrum is

jamesb93 (changes requested):

tremblap:
'confidence values. When no pitch' -> add in between: 'the 3 algorithm provided are specialised in monophonic pitch tracking and thus will struggle with anything that has 2 or more notes.


'how confident the algorithm is that there is a pitch' -> 'how confident the algorithm is on the validity of the pitch it provides' (or something like this - so we imply that it is monophonic again and dumb too)

'or “harmonic”' -> or monophonic (something along the line of stressing that a chord, very harmonic, will have a shit confidence

YinFFT: maybe we should say it is the most advance and reliable, at the cost of CPU?


note numbers (indicated by 1). - should we hint at logarithmic scale here so we seed the benefit?

src/routes/reference/pitch.svx
tedmoore: 👍 but instead of "algorithm" i'm going to say "object" since it is three algorithms.

PR #241, opened 2022-03-04
BufFlatten

tedmoore:
The widget is adapted from BufSelectEvery. Thanks @jamesb93 for your help with that! (also feel free to edit my js and svelte!)


src/lib/widget/BufFlatten.svelte
```diff
@@ -0,0 +1,176 @@
...
+    
+    if(displayNumChans===0){
+      numChans_error = 'numChans cannot = 0'
+    }
+    
+    numFrames = displayNumFrames===-1 ? srcFrames : displayNumFrames;
+    numChans = displayNumChans===-1 ? srcChans : displayNumChans;
+    
+    if(startFrame+numFrames > srcFrames){
+      startFrame_error = 'startFrame + numFrames cannot be > the number of frames in the soruce';
```
jamesb93, line 46: This isn't defined outside the reactive statement so it will throw a breaking error to the console. It also seems that it's not actually being used in the markup either I would like to go through the code anyway and normalise some of the style so I'll add this in while I'm there

src/lib/widget/BufFlatten.svelte
```diff
@@ -0,0 +1,176 @@
...
+
+<form>
+    <label class='numberboxlabel' for='start_frame'>Start Frame</label>
+    <input class='numberbox' id='start_frame' bind:value={startFrame} type=number min=0 max={srcFrames-1} />
+</form>
+
+<form>
+    <label class='numberboxlabel' for='num_frames'>Num Frames</label>
+    <input class='numberbox' id='num_frames' bind:value={displayNumFrames} type=number min=-1 max={srcFrames-startFrame} />
+    <label class="error">{ numFrames_error }</label>
```
jamesb93, line 70: `<label>` is associated to an input control in a form and this is just a warning so a `<div>` is more appropriate.

src/lib/widget/BufFlatten.svelte
```diff
@@ -0,0 +1,176 @@
...
+    <input class='numberbox' id='start_chan' bind:value={startChan} type=number min=0 max={srcChans-1}/>
+</form>
+
+<form>
+    <label class='numberboxlabel' for='num_chans'>Num Chans</label>
+    <input class='numberbox' id='num_chans' bind:value={displayNumChans} type=number min=-1 max={srcChans-startChan}/>
+    <label class="error">{ numChans_error }</label>
+</form>
+
+<form>
```
jamesb93, line 84: With your permission I think this should be an input of checkbox type (like an enum) - can I change it?

src/lib/widget/BufFlatten.svelte
```diff
@@ -0,0 +1,176 @@
...
+    }
+    
+    .error {
+        color: red;
+        font-size: 0.75rem;
+    }
+    
+    .container {
+        display: flex;
+        flex-direction: column;
```
jamesb93, line 153: I think that the layout is sufficiently small to be in rows but that is more of a thought than a request or even comment

src/lib/widget/BufFlatten.svelte
```diff
@@ -0,0 +1,176 @@
...
+        flex-direction: row;
+        gap: 0.5em;
+    }
+
+    .cube {
+        width: 30px;
+        height: 30px;
+    }
+
+    form {
```
jamesb93, line 169: There are potentially three things in the grid, so we should account for them otherwise layouts will get wonky

jamesb93 (changes requested):

src/lib/widget/BufFlatten.svelte
tedmoore: I think it should stay as a number so it is more like when a user is choosing it in a CCE.

src/lib/widget/BufFlatten.svelte
tedmoore: I don't know what most of these css attributes actually do, so you'll have to break this down for me.

src/lib/widget/BufFlatten.svelte
tedmoore: same as above: I don't know what most of these css attributes actually do, so you'll have to break this down for me.I don't know what most of these css attributes actually do, so you'll have to break this down for me.

src/lib/widget/BufFlatten.svelte
tedmoore: 👍

src/lib/widget/BufFlatten.svelte
tedmoore: oh yeah, i forgot about this guy. i ended up approaching this "warning" differently, but yes, putting it in would be good!

src/lib/widget/BufFlatten.svelte
jamesb93: Yep makes sense. Keep as is.

src/lib/widget/BufFlatten.svelte
jamesb93: You'll get in a sinch.

The `flex-direction` dictates which axis it lays out the child elements. So in this case it takes everything inside the `<div>` with the class `container` (`.class` in css) and will put it into a vertical column. You can change this to `row` to make it lay it out horizontally.

src/lib/widget/BufFlatten.svelte
jamesb93: > same as above: I don't know what most of these css attributes actually do, so you'll have to break this down for me.I don't know what most of these css attributes actually do, so you'll have to break this down for me.

All good. `display: grid` arranges child elements in a grid. the `grid-template` indicates the structure of that grid. We could make it three elements to be explicit about the structure.

jamesb93: So we can do:

`10ch auto auto` possibly. I would have to check.

tremblap:
'how audio files are' -> 'how most interleaved multichannel audio is'

otherwise this is perfect. I LOVE the widget.

jamesb93:
> 'how audio files are' -> 'how most interleaved multichannel audio is'
> 
> otherwise this is perfect. I LOVE the widget.

Then you will like BufSelectEvery too :)

jamesb93:
> > Then you will like BufSelectEvery too :)
> 
> for posterity/citation here, @jamesb93 is saying that because he designed it. i just modified it for BufFlatten.

I am saying it because we are developing a rich and enticing world based on shapes and colours. As it always should have been.

PR #253, opened 2022-03-10
[Enhance] Jargon/Wikilinks

jamesb93:
This PR adds a system for defining and automatically rendering Wikilinks inline throughout the Learn platform.

Definitions are created in `jargon.js` at the top of the repository. Any time that word is parsed out of the markdown __and__ betwixt underscores (itacilised) it will make that word hoverable and render the HTML string that is defined.

The style can be refined, but for now I think we should see how useful this is and we can make it all pretty later...

Reviewers do you mind having a look at this briefly and providing feedback from the perspective of developing with it, as well as how it looks for readers?

For example, see the OnsetSlice entry's first paragraph.

@tedmoore @jdchart


jdchart:
Oops looked at this when you posted yesterday but forgot to comment.

Think it looks great! perhpas the style of the inline link needs to be made a slight bit more apparent but the style of the qctual popup i find good !

tedmoore (approved):

PR #258, opened 2022-03-15
[Enhance] Expand/Contract Search

jamesb93:
Makes the search expand/contract on focus﻿


jamesb93:
Think this looks a bit naff now. Going to close.

PR #260, opened 2022-03-15
[Content] Snippets

jdchart:
for the moment, the system works thusly:

- There's a folder in static called snippets, within which each snippet has its own folder that contains 4 files: max.txt, sc.txt, pd.txt and meta_data.json (title, video url, blurb etc). So very easy to add/maintain snippets.
- On build the site now builds a snippet db, parsing all that crap.
- app.js now finds the snippet matching the "featured_snippet" in the manual_config file, which is just the name of the snippet's folder.
- this featured snippet data get sent to index for display. But there's also scope for showing more if we want.

Obviously its a bit fugly atm, but there we are. @jamesb93 and @tedmoore , would love to hear what you think.


jdchart:
Note/question james, i figured out that this is how you run scripts on build:

```
"dev": "node scripts/metadata.js && node scripts/db.js && node scripts/related.js && node scripts/snippet_parse.js && svelte-kit dev --host",
"build": "node scripts/metadata.js && node scripts/db.js && node scripts/related.js && node scripts/snippet_parse.js && svelte-kit build",
```

Is there perhaps scope/possibility for just giving a master script here that runs the others? or am i sadly deluded?

PR #270, opened 2022-03-25
[enhancement] toc adjustents

jdchart:
Was going off screen on my smaller screen.


jdchart:
@jamesb93 i think the js is perhaps janky, but anything involving stuff that happens on scrolling and screen resizing naturally will be. I think it works well.

PR #271, opened 2022-03-25
BufStats

tedmoore:
As BufStats is one of the *most* used objects in the toolkit, and is quite powerful, this reference page is a bit of a doozy. I've already re-thunk a few of these sections a number of times and there's some content up here I think maybe shouldn't be, but I've left it all as fodder for us. Looking for feedback from all!

Some of the questions I've been considering:

* I'm thinking that some of these sections will end up as separate Learn pages (you'll notice the "Weights" section already points to a Learn page that doesn't exist yet).
* I think sometimes I explain something with two sentences that maybe should instead be a jargon scroll over popup?
* Where can there be more musicianly examples sprinkled in here?
* I'm trying to make sure that these different facets are presented as *understandable* concepts and we don't rely on "trust me" types of descriptions or the overuse of metaphors (although of course they can be helpful!), or calling things "magic" which basically says to the reader "it's too complicated for you"--which I think means they won't really try to use that feature of the object.
* I'm also trying to make sure these things feel *useful*. Without a clear description of how `outliersCutoff` really works I just don't think people will try to use it. I'm hoping all these descriptions can make a reader think (1) "I understand what that thing does" and (2) "That seems like that would be useful for a situation like _x_"

(also there are some formatting things and image quality issue to fix, but that can happen after some content revision)


weefuzzy:
## Structural

Overall, it needs to be shorter. Big suggestions in this respect are to make stuff about distributions / histograms its own article (relevant also for *many* other objects) and try to condense the outliers example a bit (or consider making that its own thing, as it's pretty involved). 

## Blow by blow 

> returning seven statistical properties

-> returning seven statistics 

> while the others offer a measure of the variability in the data.

Ho hum. This is pretty imprecise. How about 'the others help describe how the data are distributed around this tendency'? Or something. 

* Paragraph 2: swap the clauses so that you lead with the proposition that statistical summaries can be useful for comparing the properties of different length time series? 

> In addition... 

Not sure about dropping this bomb so early (i.e. before dealing with the base quantities) 

> Standard deviation describes of the amount of variation in the data ...

*in relation to the mean* (which implies that it only has much use when the mean itself is telling us something useful, which isn't always) 

Following para discussing interpretation w/r/t Normal Distr: my concern here is that people will read this without caveats and start to apply it blindly. It *can* be useful of course, e.g using a rule-of-thumb 3σ threshold for detecting something interesting can be a useful trick. But: I'm concerned that the relative weight of this paragraph end up over emphasising a set of conditions unlikely to be met in practice. 

The figure of the normal distribution appears twice in the article. I think for this first time, having all the annotations about amounts of data under the curve is distracting (our reader still possibly doesn't know what a distribution  is, so still has to come to terms with what such a plot represents). Also the x axis labels assume this a 0-mean distribution, without justification, and the y-axis labels aren't informative. So, maybe – if anything– just something that shows the bell shape at this point? 

> Skewness indicates asymmetry in the [distribution](https://learn-git-bufstats-flucoma.vercel.app/reference/bufstats/#distribution-and-histograms) of the data.

*in relation to the mean* 

kerning on the figure's axis labels is all messed up 

> Kurtosis describes the degree to which outliers are present in the data

Careful. We haven't described what an outlier is (and it's a slippery concept). Again, this is in relation to the mean, and whether or not we can really call something an outlier hinges on whether the arithmetic mean tells us something about the *expected value* for the data / distribution in question. If Gaussian(ish), then sure. If uniform, then not so much. So, it tells us about the 'tails' of our observations or distribution around the mean.

I think it's worth taking care here, because when we come to discuss outliers later, it's in terms of observations that can distort our estimates of things like mean (because they're presumed to be in violation of some – more or less justified – assumption about a 'true' underlying distribution) 

* Low, Middle, High, Percentiles 

The ordering of concepts feels funny here, as the descriptions for low / middle / high all depend on the notion of a percentile. It might be worth introducing them together as Order Statistics, and emphasising up front that they're quite independent from the preceding quantities and can be more robust in dealing with data that are noisy, or strongly non-Gaussian. 

> When the quantity of values does not make clear which point from the data should be used as a percentile, a [nearest rank method](https://en.wikipedia.org/wiki/Percentile) is used, which ensures that the percentile values returned are always values actually found in the data.

This isn't very clear to a novice, I don't think. A useful motivating example of why you might want measures that return actual observation values might be something like pitch, where the mean would quite likely return pitches that are, uh, spicy in relation to the sonority at hand. 

> the derivative of a time-series is computed by finding the difference between each consecutive value in a series.

computed -> approximated (this is significant: difference function approximations of derivatives are by their nature noisy and may be wholly unreliable). Not wild about the spreadsheet stuff. Can we present the motivation up front and jump straight to the example? 

### Weights

Difficult to describe the usefulness without reference to outliers as an idea, so maybe this comes after? Motivating example could be to use pitch confidence as a weighting on estimated pitch, so that where the algo is less certain, these measurements have less influence on the estimated statistics. 

### Outliers 

I think you need to bite the bullet and say what an 'outlier' is for these purposes (and it's necessarily subjective), and why they can be problematic. Given that we also need this concept to explain why `RobustScaler` exists, this is another candidate for splitting out. I don't think the data tables are very appealing as an explanation: something graphical would be easier to grasp, IMO 

### Distribution and Histogram 

Say how we might interpret (e.g. the probability that any observation in our data sample takes on the value that it does, leaving aside all the nuances of what we mean by probability etc). [oh, I see you get to that in the following para. move it up?]

> A distribution is often viewed as a histogram
 
approximated as a histogram 

> Sometimes the distribution is represented as a curve which can be called a distribution function. This can be thought of as a smoothed out version of a histogram.

It really shouldn't be thought of that way. Histograms are unhelpfully dependent on the choice of bin width, so can end up looking markedly different to a PDF estimated some other way. 

If this is to be separated off, then I think we can introduce a uniform distribution as well, which affords a simple interpretation and is also pretty common. It might help with making things clear. 

> Normal distribution is a technical term meaning that the distribution of data follows the 68/95/99.7 rule. 

This is certainly not all that it means 😬 e..g it also means that your mean, median and mode are coincident; that the mean and standard deviation tell you all you need to know about the distribution of data; and that a bunch of convenient mathematical properties hold true. Importantly (IMO), just as the FFT gets used as a TF representation so often because it is well understood and easy to work numerically with, rather than being in any sense The Best Thing, likewise Gaussian assumptions are extremely common in ML models because they're well understood and easy to work with. 


## Anything missing? 

Somewhere I think we need to point out how common Gaussian and i.i.d. assumptions are to the various algorithms we provide, but that they're never really true (time series are typically not i.i.d, and often non-Gaussian, for example) which amounts to a big fat YMMV when trying to interpret these numbers.

tedmoore:
Thanks @weefuzzy this is really helpful! You've struck on some of the things I was also feeling were poor or slippery and I like your suggestions. I'll incorporate this stuff and see what follow on questions I have.

(I'll wait for others to chime in too though)

tremblap:
I've seen @weefuzzy thorough feedback, which I will read after. After reading the first half of the article, I think we need to split it, have more qualitative description, a very big caveat about distribution assumptions and how this is limited in 'accuracy' even in the qualitative bit, and the musical example before the maths.

I think that loads of small articles could be easier to maintain (and easier for our readership to see what is overview and what is detail). the sections I see
- qualitative description of the object (seeding mean-related (first four) and centile-related (last 3)) and the caveat and a musical example
- the explanation of mean-related as an article
- the explanation of centile-related as an article
- outliers as an article (with a big fat std assumptions again although it helps a bit to alleviate)
- weighing article

I think you have done a very good job of putting it all to the page. we can talk order. Now I'll read Owen's feedback to see if I agree with it all as I do most time, or propose a nuance or counter opinion :)

tremblap:
ok reading what Owen has so well articulated, I agree. many of the order ideas and focus and caveats are similar. a few more things maybe to consider:

- when we talk about std assumptions (I live the 'easy to understand' explanation) I think we should show a bimodal asymetical graph and show how it fails. then point at Khan academy for people who really want to understand the depth of the problem.

- I had to check in the source code, I didn't know we capped derivatives at 2. There is no reason (apart from limited returns) so maybe we shouldn't stress it (it is written more than once, as if it was important). or we explain why it is becoming less and less relevant and more abstract.

- kurtosis - I like spikiness as a metaphor. it is also true if not centred on the mean. I remember Gerard giving me a lot of cool explanations and examples. and Peteers did a sort of good musicianly explanation of these 4 (r.e. centroid but still)

congrats again for that mega endeavour

tedmoore:
> I think that loads of small articles could be easier to maintain (and easier for our readership to see what is overview and what is detail). the sections I see
> 
> * qualitative description of the object (seeding mean-related (first four) and centile-related (last 3)) and the caveat and a musical example
> * the explanation of mean-related as an article
> * the explanation of centile-related as an article
> * outliers as an article (with a big fat std assumptions again although it helps a bit to alleviate)
> * weighing article

I'm glad this is coming to the fore. I agree. The breakdown you have here is in the realm of granularity I was thinking as well.

tedmoore:
@tremblap @weefuzzy @jamesb93 

I *believe* 🤞 I have incorporated all the feedback. There are a few things I may have *not* incorporated, for which I probably have a reason, which I'd be happy to share...(hopefully it's a good reason!)

There are many new pages spawned from the original (which maybe should still be shorter...), so check out:
* BufStats (the original)
* Distribution and Histograms
* Outliers
* Removing Outliers
* Weighting Stats

Also there's a page called "Statistical Assumptions" which is not complete yet. It currently has a paste of a couple of ideas from the previous feedback. I'm hoping that you all can provide some ideas, resources, (prose!?), etc. that can go in there and then I'll take a whack at it. My gut right now is that it should be a couple paragraphs of broad ideas (and/or warnings) and then point to further resources, rather than trying to encompass the whole of statistical fallacies and pitfalls.

All feedback desired!

//================================================================

One response to @weefuzzy's comment regarding the data tables in the (old) outliers section:

> I don't think the data tables are very appealing as an explanation: something graphical would be easier to grasp, IMO

I agree and I think on the new "Outliers" page there should be a more graphical representation as you suggest to ponder the concept of "outlier". If anyone has a nice idea for this I'd love to hear it. I want it to be musically (and FluCoMa-y) relevant and not just [something off of wikipedia about the speed of light experiments](https://en.wikipedia.org/wiki/Outlier#/media/File:Michelsonmorley-boxplot.svg). Also I want it to be not really contrived, more like something that a user would find "in the wild". And also hopefully something that is an outlier which would be reasonable to throw out, so there's some more motivation and context around the idea (not just, "look, these dots are far away" because those could just be really different sounds, not "outliers", yes the whole concept of outlier is slippery). I tried 2D plotting the trumpet example on the "Outliers" page but it looked too contrived.

The thing I'm hoping these data tables *can* show is how the `outliersCutoff` functionality operates in a specific and technical way. I want to be sure to de-magic-ify how this works, because I think if it's magical it means people don't really understand and if they don't really understand they'll be hesitant to use it (i.e., never will). So...I've moved these data tables to the Learn page that is specifically for explaining how the `outliersCutoff` functionality works because I think they're good for that. That being said, any further or different ideas about how to best convey the functionality are very welcome.

weefuzzy:
> I agree and I think on the new "Outliers" page there should be a more graphical representation as you suggest to ponder the concept of "outlier". If anyone has a nice idea for this I'd love to hear it

Spectrogram with dropouts: zero-energy frames are clearly (enough) outliers from the rest of the action. Likewise, spectrum of mostly steady material interspersed with sharp transients. 

I think it's important to make clear the subjective quality to outlier-hood: one has to decide that outliers are problematic in some concrete context and *then* decide what constitutes an outlier in that context. Starting with things like > 3σ rules just lead to weird circularity (because σ would be different without the outliers). Both suggestions above give a chance to emphasise that subjectivity. 

> Also there's a page called "Statistical Assumptions" which is not complete yet. It currently has a paste of a couple of ideas from the previous feedback. I'm hoping that you all can provide some ideas, resources, (prose!?), etc. that can go in there and then I'll take a whack at it. My gut right now is that it should be a couple paragraphs of broad ideas (and/or warnings) and then point to further resources, rather than trying to encompass the whole of statistical fallacies and pitfalls.

I'll have a look at this once it's deployed?

jamesb93:
Issue was upstream in SvelteKit so I've fixed the version to 321 for that dependency. I've also resolved the conflict this created in `package.json`.

jamesb93:
On the content itself, I think that this is much nicer in terms of opening up the idea of statistics to interested peoples and reducing the monolithic entry that was there before. It's also inviting for future additions to this sphere (like the fallacies one to come).


In terms of outliers, I think tapping into the example @weefuzzy mentioned will be powerful and perhaps could be executed with a bit of `DataSetQuery` to demonstrate how it can be managed in the FluCoMa-verse. In addition to that there is a link here to @tremblap's LPT example, which uses a custom set of "replacement" descriptor values to try and curate analysis windows that have funky values (extremely low confidence/loudness etc). This is another approach and tie in.

On a final note, there are a few tidbits to work through here. Each article needs a "featured image" for the display on the front page. They can be whatever you want but have to be specified in the frontmatter of each `.svx`. For an example see here:

https://github.com/flucoma/learn-website/blob/d587fc0f802c26978062109e1b7c45e9f0010cef/src/routes/learn/bufnmf.svx#L11

There are also inconsistencies in capitalisation of headers. For example: 

"Might they be interesting"

vs.

"Working with Outliers"

I have been silently following the convention proposed in the [APA style](https://blog.apastyle.org/apastyle/2012/03/title-case-and-sentence-case-capitalization-in-apa-style.html).

jargon.js
```diff
@@ -3,9 +3,11 @@
...
+    mean: "The average value of the data. This is calculated by adding up all the numbers and the dividing by how many numbers there are.",
+    median: "The value at or below which 50% of the values in the data will fall, also called the 50th percentile.",
+    mode: "The value that occurs most frequently in a set of data."
 }
 
-
-
-export default { definitions };
\ No newline at end of file
+export default { definitions };
```
jamesb93, line 13: LOVE IT

src/routes/learn/distribution.svx
```diff
@@ -0,0 +1,176 @@
...
+    - standard deviation
+    - mean
+    - median
+    - mode
+    - normal
+    - histogram
+    - stats
+    - statistics
+flair: tutorial
+---
```
jamesb93, line 16: featured image needs to be specified here.

see https://github.com/flucoma/learn-website/blob/d587fc0f802c26978062109e1b7c45e9f0010cef/src/routes/learn/bufnmf.svx#L11 for an example

src/routes/learn/outliers.svx
```diff
@@ -0,0 +1,72 @@
...
+    - normal
+    - histogram
+flair: tutorial
+---
+
+<script>
+    import Image from '$lib/components/Image.svelte';
+</script>
+
+## What is an outlier?
```
jamesb93, line 18: Header capitalisation inconsistency.

To reduce noise I'll just flag this once here.

src/routes/learn/removing-outliers.svx
```diff
@@ -0,0 +1,220 @@
...
+blurb: |
+    Prevent outliers from negatively impacting BufStats' statistical summary.
+tags: 
+    - mean
+    - median
+    - mode
+    - normal
+    - outlier
+flair: tutorial
+---
```
jamesb93, line 12: featured image

jamesb93 (changes requested):
1. Header capitalisation
2. Featured Images for each learn entry.

PR #272, opened 2022-03-25
Reference: MFCC

tedmoore:


tremblap:
the preview of this fails, should I wait to give my 'review'?

tremblap (commented):
this is amazing. a little change of order would make it clearer for me

- the green box would go at the very bottom and probably not as a green box, and flagging "for a more thorough exploration of the impact of each coeff" or somehting like this. it is amazing after all so we should brag :)

- the widget in the 'main' page would go between the last 2 paragraph - instead of at the bottom. so the explanation on the blue line is under the widget.

great work!

tremblap:
btw on the in-depth page, when I click on a coeff from a 'real' analysis the whole thing changes, is that normal?

PR #277, opened 2022-03-29
Update median-filters.svx

Mcas4150:
duplicate sentence here


jamesb93:
This is great – thanks for these small but meaningful changes @Mcas4150.

PR #280, opened 2022-03-31
[Enhance] More compact related links section

jamesb93:
A response to: https://github.com/flucoma/learn-website/pull/270

Check out the [Harker](https://learn-git-fork-jamesb93-hack-toc-layout-flucoma.vercel.app/explore/harker/).

On another note it completely shafts other pages like [HPSS]().

image

I think that this is symptomatic of a bigger problem, the pages are inconsistently structured. For example there is no h1 on the HPSS page which leads to the kind of strange solitary "median filter" tag there. Not to mention HPSS is mentioned in many other places but this doesn't get captured in that... There needs to be both related _forward links_ (what we have now) and _backlinks_. For example HPSS is linked a lot on Lauren Sarah Hayes article... noone would know that if they came there!


jdchart:
> On another note it completely shafts other pages like HPSS.

I think completely shafted may be an overstatement. 

> There needs to be both related _forward links_

This is true, probably not too difficult to implement with some parsing...

I think this is globally a lot better ! And, assuming there isn't an article that's hundreds of H2s long, should clear up the problem of the stuff on the side going out of smaller screens! (and also makes things cleared up and tidier).

jamesb93:
> I think completely shafted may be an overstatement. Just a bigger lower margin may do the trick.

yeah true...

Another option:
image

> This is true, probably not too difficult to implement with some parsing...

What we had prior to this PR are forward links :)

> I think this is globally a lot better ! And, assuming there isn't an article that's hundreds of H2s long, should clear up the problem of the stuff on the side going out of smaller screens! (and also makes things cleared up and tidier).

cool, thanks for the feedback :)

PR #286, opened 2022-04-06
[Layout] content index mosaic

jdchart:
@jamesb93 you mentioned wanting this, what do you think ?


jamesb93:
Looks really good. Might need some tweaking but I don't quite have the headspace to figure out what the tweaks are right now... I think it is kind of dense to look at from a glance? Maybe try going a bit wider on this layout and seeing if more padding between items helps.

jdchart:
> Looks really good. Might need some tweaking but I don't quite have the headspace to figure out what the tweaks are right now... I think it is kind of dense to look at from a glance? Maybe try going a bit wider on this layout and seeing if more padding between items helps.

Cool, I'll mess around with some stuff

jamesb93:
I think we'll need something like this:

https://svelte-bricks.netlify.app

PR #304, opened 2022-05-05
MLP Reference

tedmoore:
Feedback desired. There are three pages:

mlp-training
mlp-parameters
training-testing-split

As with stats there's a few things that I am still a little squishy about but I leave for your fodder.


src/routes/reference/mlpclassifier.svx
```diff
@@ -13,36 +13,16 @@ flair: reference
 category: Analyse Data
 ---
 
-The `MLPClassifier` is an object that can be used to perform _classification_. In machine learning, _classification_ refers to an algorithm that is trained to predict what category, or _class_, an input belongs to. 'MLP' stands for multi-layer perceptron which is a type of neural network.
+<script>
```
jamesb93, line 16: I find these pages neat but a little "compact" in that information is not enumerated very clearly. Like... which order should I take these in? etc... I think that anywhere there is a video we could be using richer content and components such as <YouTube />.

jamesb93: I also think a generalisable graphic might help be an evocative entry point versus quite a lot of white space and minimal text.

src/routes/reference/mlpregressor.svx
```diff
@@ -13,39 +13,16 @@ flair: reference
 category: Analyse Data
 ---
 
-The `MLPRegressor` is an object that can be used to perform _regression_. In machine learning, _regression_ can be thought of as a mapping from one space to another, much like mapping a controller onto synthesis parameters. 'MLP' stands for multi-layer perceptron which is a type of neural network.
+<script>
```
jamesb93, line 16: Same feedback as the `mlpclassifier.svx`

src/routes/learn/mlp-training.svx
```diff
@@ -0,0 +1,232 @@
...
+label="A multi-layer perceptron neural network that takes two inputs, has one hidden layer of seven neurons, and predicts ten outputs. Each black arrow is a weight. Each of the neurons in the hidden layer and output layer also have a bias."
+/>
+
+### Audio Mixer Metaphor
+
+A useful metaphor for understanding how the feedforward algorithm works is to think of a neural network like an audio mixer. The inputs to the neural network are like the audio inputs to the mixer. The weights are like volume adjustments that control how much of an input signal gets passed on (in the case of a neural network, passed on to the next layer). The biases are like [DC offsets](https://manual.audacityteam.org/man/dc_offset.html#:~:text=DC%20offset%20is%20a%20mean,and%20how%20to%20remove%20it.) that just shift the signal up or down. [Activation functions](/#activation) are like distortion--they introduce some non-linearities into the signal (which for neural networks help them learn non-linear relationships between the inputs and outputs). Lastly, the outputs of the neural network are like the outputs of the audio mixer--a volume adjusted, DC offset, and distorted combination of all the inputs.
```
jamesb93, line 52: `[Activation functions](/#activation) are like distortion--they introduce some non-linearities into the signal (which for neural networks help them learn non-linear relationships between the inputs and outputs).`

Needs a space and I wouldn't use -- but instead – (and not -) which are distinct.

src/routes/learn/training-testing-split.svx
```diff
@@ -0,0 +1,34 @@
...
+
+### Overfitting
+
+Using a testing set can help determine if a neural network _model_ is overfitting. Overfitting means that the neural network has learned to _really accurately_ make predictions on the data it was trained with, but as soon as it is shown data it has never seen before it performs quite poorly. Another way of stating this is that the moodel cannot _generalise_ to data it hasn't seen before. If a model is able to perform well on both the training data and the testing data, it is likely not overfit. If a model is able to perform well on the training data but does poorly on the testing data that was split off, it is probably overfit to the training data. In this case, there are few thing to try to prevent overfitting:
```
jamesb93, line 26: moodel > model

src/routes/learn/mlp-training.svx
```diff
@@ -0,0 +1,232 @@
...
+
+### The Error
+
+Because training the neural network is an iterative process, it is likely that you'll send the `fit` message to the neural network many times. Each additional time it receives `fit`, it continues training, picking up from wherever it left off (the internal state of the neural network is not reset). Each time it concludes a fitting it will return a number called the error (also known as the loss). The error is a measure of how inaccurate the neural network still is at performing the task you've asked it to learn (so it may need more training!). Generally speaking, lower is better because it means there is _less_ error, therefore the neural network is more accurately performing the task it is learning.
```
jamesb93, line 62: Perhaps this is missing one part of the puzzle. The larger your network the bigger your error values _can_ be (MSE etc...) So you can infer how good your error is if you normalise for the size of your network which we don't do for free.

src/routes/learn/mlp-training.svx
```diff
@@ -0,0 +1,232 @@
...
+
+One strategy of training the neural network is to set [`maxIter`](/learn/mlp-parameters#maxiter) rather low (between 10 and 50 maybe) and repeatedly call `fit` on the neural network. In Max or Pure Data, this might be using a [metro] (or [qmetro]). In SuperCollider this could be done by creating a function that recursively calls `fit` unless a flag is set to `false` (see the example code below). While these sequential `fit`s are happening, watch how the error changes over time and adjust some of the MLP parameters _as it is training_ to see how they affect the change and rate of change of the error value. Finding which parameters create a good training (_convergence_ at a low error) is often a process of testing various combinations of parameters to see what works best. All of the parameters are worth experimenting with! And as always, make sure to [test the performance](/learn/training-testing-split) of the neural network _model_ to see if it really does what you hope!
+
+<Tabs>
```
jamesb93, line 74: needs a pd example

src/routes/learn/training-testing-split.svx
```diff
@@ -0,0 +1,34 @@
...
+---
+
+### Training Data and Testing Data
+
+Before training a neural network it may be important to separate the input-output example pairs into a **training set** and **testing set**. The **training set** will be used to do the actual training of the neural network and the **testing set** will be used after the training to test how well the _model_ is performing. This can be important to simulate how the neural network might perform in the actual use-case you're aiming for. For example, if you [train an MLPClassifier to distinguish oboe and trombone timbres](/learn/classification-neural-network) so that during a live performance it can know which instrument is sounding, you would train on a **training set** of audio analyses that have corresponding labels: either "trombone" or "oboe". Then, to test the effectiveness of the training, you would make predictions using a **testing set** that has both oboe and trombone examples which were not part of the training set. This simulates the future use-case: having the _model_ make predictions when a trombone and oboe are on stage making sounds, the analyses of which it has never seen before.
```
jamesb93, line 18: maybe instead of **bold** you could use _underscores_ and put a jargon entry?

src/routes/learn/training-testing-split.svx
```diff
@@ -0,0 +1,34 @@
...
+
+* If you have enough data such that it makes sense, use a [`validation`](/learn/mlp-parameters#validation) set.
+* Try training the neural network to a not-as-low loss value. This may stop it from reaching a state of overfitting.
+* Decrease the number of internal parameters in your neural network by having fewer hidden layers and/or fewer neurons per hidden layer. Neural networks with many internal parameters have the potential to overfit because they are able to essentially "store" the training data's specific input-output relationships within their many internal parameters. Having fewer internal parameters can prevent this from happening, instead forcing the neural network to learn a more generalised understanding of the input-output relationships. The more generalised understanding will perform better on the testing set (and other data that it hasn't seen before).
+
+### Underfitting 
+
+Overfitting can be contrasted with underfitting in which a _model_ also cannot _generalise_, in this case because it probably has not been trained enough. A model that performs poorly on the testing set, but also performs poorly on the training set is likely underfit.
```
jamesb93, line 34: underfit should be underfitted? I think...?

jamesb93 (changes requested):
This suite of articles is good. Adds a lot of rich content to an otherwise bloated set of resources out there on the internet and answers many questions from the workshops.

I have a few layout / spelling issues flagged as comments. Happy to clarify and talk more about them if its needed but I trust your judgement otherwise.

src/routes/learn/mlp-training.svx
tedmoore: @tremblap, I think we should add this to Martin's list?

tremblap (changes requested):
also, hidden is now hiddenLayers in our objects

tedmoore:
> I checked the pages reference/MLPClassifier and reference/MLPRegressor. nice and short. In both cases I wouldn't point to the youtube tutorial directly, I would point to the tutorial pages (learn/classification-neural-network and learn/regression-neural-network/)

Good, yes, I agree.

> I would put comparing knn and mlp at the end of the page

👍 

> (sad there isn't pd patches - we should put them there too)

Can we have Martin do these translations as well?

> we could hint there at representative sampling right? an external reference for the mathematically enclined?

Good idea. Do you have an external reference in mind?

> when i click on more info below I arrive back to the start page... is that a bug?

Yes it is!

> the menu on the left adds inlinecode,hidden, inlinecode,learnRate etc is that a bug?

Yep. @jamesb93 and I have discussed and are on the case.

> that article explains the parameters and ways of training - isn't that confusing to have that here and in mlp-training? 

Can you be more specific (like down to the sentence?).

> I think this is more technical than mlp-training so maybe we should clarify that at the top of the page - the other is the overview of training, and this is about tweaking and tricks around parameter, IIUC?

Yeah, this more technical. I kind of think about the MLP Training page as the 101 information that we give in workshops...and then everyone says that they want to know _more_ about the parameters, so that's where the MLP Parameters page comes in. However, I do want them to feel like they can stand alone as well as I think they will have different uses for different people at different times, so even if there is some overlap, I think that's ok if it makes for pages that feel more cohesive unto themselves.

> in learnrate: "Then after a few fittings, decrease the learning rate to a smaller value, maybe 0.01, to slow down" - maybe we could talk about when it starts to bounce like we spoke about on discord

Cool. Good suggestion.

PR #314, opened 2022-05-11
learn/fourier-transform and ref/bufstft

tedmoore:
@tremblap and I had talked about an example where a square wave is processed using BufSTFT so that we could see how changing the phases will greatly change the waveform, but not so much change the magnitudes. I hacked at this for a while (https://github.com/flucoma/graphics/tree/main/stft) but I can't figure out a narrative for it that feels really relevant to a use case or something, so I've left it out for now.


tremblap:
this is good

fourier-transform:
strange bold of wondwosize / 2 - also, spell out divided by 2
must be >= -> spell out greater or equal than - in the following sentence > again spelt out

i like the piano - bins map. it would be even better if we could swap the 2 views - i.e. compress the piano keyboard - no?

I kind of miss my blurry snapshots of iZotope on Nicol - they show yet another way to get an intuition of smearing (before sending people to the technical circle thing you say is great. That would wrap up that section quite well (on the interaciton between time and spectral resolution)

(linear vs. piano) in linear frequency is that a missing image?

very good list of references

===
bufstft

remove the (almost) in the sentence about perfect reconstruction. It is perfect with nothing is touched.

after the hann window condition for perfect reconstruction sentence, you could say 'hence the default of -1 giving an overlap of 2' to bat the take-home-message home

typo: one magnitude of phase (magnitude or phase)

typo: moroe -> more?

an audio/visual example of smearing and pre-ring would be great after the verbal description. I think a bass drum is a good source for that with low-end content and sharp transient.

good work!

tedmoore:
> i like the piano - bins map. it would be even better if we could swap the 2 views - i.e. compress the piano keyboard - no?

Totally. Do you know of an existing image that shows this? I can create that one too--just takes a little doing and I have some other things that are perhaps a higher priority. It would probably just be MIDI notes, instead of a _keyboard_ because drawing a keyboard that is stretched to specific non-linear proportions, sounds quite fiddly.

> I kind of miss my blurry snapshots of iZotope on Nicol - they show yet another way to get an intuition of smearing (before sending people to the technical circle thing you say is great. That would wrap up that section quite well (on the interaciton between time and spectral resolution)

Yeah, I can put this in (do you those pictures? so I don't have to regenerate them?).

> (linear vs. piano) in linear frequency is that a missing image?

Yeah, that's where the stretched piano would go if/when it gets added. (forgot to remove my placeholder!)

> remove the (almost) in the sentence about perfect reconstruction. It is perfect with nothing is touched.
> 
> after the hann window condition for perfect reconstruction sentence, you could say 'hence the default of -1 giving an overlap of 2' to bat the take-home-message home

The "almost" was a hold-over from the original version I had started from. I didn't want to remove any important qualifications! But yes, I'll remove the almost.

> an audio/visual example of smearing and pre-ring would be great after the verbal description. I think a bass drum is a good source for that with low-end content and sharp transient.

For sure. If you have the image and audio cued up that'd be great!

tremblap:
> Totally. Do you know of an existing image that shows this?

nope sadly

> do you those pictures?

attached - titles have the settings on

>  audio cued up 

the first 3 secs of Nicol


fft-1024-lin

waveform
fft-1024-mel
explana
fft-8192-lin
fft-1024-log

tedmoore:
> titles have the settings on

Unfortunately github strips those on upload, can you label them!?

tremblap:
in order:
1024 lin
pcm
1024 mel
8192 lin
1024 log

PR #316, opened 2022-05-23
PCA Reference

tedmoore:


src/routes/reference/pca.svx
```diff
@@ -12,4 +12,45 @@ flair: reference
...
+
+## PCA for Dimensionality Reduction
+
+Because PCA orders the principal components from those that demonstrate the most variance in the data to the least, one can effectively perform dimensionality reduction by removing some of the higher ordered principal components (in the transformed data set) because they are less useful at representing the variance in the data. This allows the data set to have fewer dimensions, while retaining the aspects of the data that are more useful in understanding the differences between the data points (the lower order principal components).
```
jamesb93, line 34: Reading so far there is a lot of caveated bracketed writing which I find bumpy. I think it doesn't need to be expressed in this way. For example:

> In order to calculate the principal components, PCA analyses the original DataSet and determines how to "rotate" the data so that an axis can be drawn through the data along which the most variance in the data can be observed (for a excellent visual explanation, see the YouTube video above)

This is just two sentences. Furthermore we should be linking to a timestamp.

> PCA continues the process of "drawing axes", through the data (always maximising variance) until all the principal components have been determined (PCA computes as many principal components as there are dimensions in the original DataSet)

PCA continues the process of "drawing axes" through the data until all the principal components have been determined. It consistently aims to maximise the variance of these axes and will compute as many principal components as there are dimensions in the original DataSet.

> This process orders the principal components (starting at 1) according to how much of the variance in the DataSet they are able to demonstrate (from demonstrating the most variance to the least)

This process orders the principal components according to how much of the variance in the DataSet they are able to demonstrate. In other words, the first principal component demonstrates the most amount compared to the last.

src/routes/reference/pca.svx
```diff
@@ -12,4 +12,45 @@ flair: reference
...
+
+As this is often how PCA is used, FluidPCA allows the user to specify the number of output dimensions desired, so that `transform` operates as a dimensionality reduction algorithm (other dimensionality reduction algorithms in FluCoMa include [UMAP](/reference/umap) and [MDS](/reference/mds). After FluidPCA transforms a DataSet it also returns the _explained variance_: the fraction (between 0 and 1) of the total variance in the DataSet accounted for using the reduced number of dimensions it has transformed the DataSet into. Values close to 1 mean that most of the variance in the original data is represented while number near 0 mean most of the variance in the original data is not represented in the transformed DataSet. When using PCA, knowing the explained variance is quite important. Reducing the number of dimensions while loosing most of the variance will likely be problematic further in the processes. Being able to reduce the number of dimensions while retaining most of the variance can help avoid the curse of dimensionality, speed up future computations, and/or remove redundancy or noise in the data. Aiming to retain an explained variance of 0.99, 0.95, or 0.9 are common.
```
jamesb93, line 36: IMO would be better raised as a <Note type="pointer">

jamesb93: Apologies for lack of specificity here: the pointing toward the other dimension reduction algorithms.

jamesb93: FluidPCA should just be PCA>

jamesb93: Loosing > Losing

src/routes/reference/pca.svx
```diff
@@ -12,4 +12,45 @@ flair: reference
...
+
+## `whiten`
+
+PCA "whitening" can be turned on by setting the `whiten` parameter to 1. "Whitening" ensures that not only will the principal components be uncorrelated, but also they will all have unit variance (they'll all cover the same range of 1). "Whitening" is sometimes also referred to as `sphering` meaning that it's making the data "spherical" in the sense that each principal component (each new dimension) has the same spread, just as a sphere has the same width in each of its three dimensions.
+
+Whitening looses some information during the transformation process (essentially, the relative scales of the principal components), however, whitening may improve the accuracy of further computations such as classification. If there is noise in the data, whitening might amplify this noise by scaling it to be the same size as other other information, making it as prominent as the relevant information, therefore whitening will not _always_ improve data.
```
jamesb93, line 42: It is not whitening that loses information but the transformation itself that loses it from the original data so I think the sentence needs to be re-articulated. Or, you could say whitening causes a loss of information.

src/routes/reference/pca.svx
```diff
@@ -12,4 +12,45 @@ flair: reference
...
+<ResourceLink
+title='Principal Component Analysis (PCA) - Computerphile'
+url='https://www.youtube.com/watch?v=TJdH6rPA-TI'
+blurb='YouTube video containing great visual explanations of PCA'
+/>
+
+<ResourceLink
+title='PCA Whitening'
+url='http://ufldl.stanford.edu/tutorial/unsupervised/PCAWhitening/'
+blurb='Some of the maths behind PCA Whitening.'
```
jamesb93, line 55: is Whitening here a proper noun?

jamesb93 (changes requested):
I like it and think the content is robust. I have a small issue with the overuse of brackets to express caveats, which makes it seem more complicated and hard to track as a reader.

src/routes/reference/pca.svx
tedmoore: I don't understand what you're asking? Are you suggesting that whitening should be lowercase?

src/routes/reference/pca.svx
tedmoore: > Apologies for lack of specificity here: the pointing toward the other dimension reduction algorithms.

where is this refering to?

tedmoore: > FluidPCA should just be PCA

I'm trying to make a distinction between the algorithm and the object. It wouldn't be true to say that

`PCA allows the user to specify the number of output dimensions desired`

but it is true to say that about the object.

src/routes/reference/pca.svx
jamesb93: If it is not a proper noun then yeah

PR #318, opened 2022-05-24
Standardize

tedmoore:


tremblap:
this is super clear. I think we should add a caveat that our real-life audio data is very rarely standard...

tremblap:
> this is super clear. I think we should add a caveat that our real-life audio data is very rarely standard...

referencing the cool page on distribution there could be good actually with that caveat.

jamesb93 (approved):
PAs suggestions are good and I have nothing to add.

PR #319, opened 2022-05-25
comparing scalers & why scale?

tedmoore:
This introduces two new "learn" pages: "Why Scale? Distance as Similarity" and "Comparing Scalers"


jamesb93 (changes requested):
One small catch on KMEANS is all I have to ask for changes.

I think that this could be better supported with some interactive widgets in the future but thats a nice to have. I'm going to make an issue to loop around to it because I think, in particular, the first graph could be made better by being something that morphs as you alter the scales or something.

tedmoore:
> I think that this could be better supported with some interactive widgets in the future but thats a nice to have. I'm going to make an issue to loop around to it because I think, in particular, the first graph could be made better by being something that morphs as you alter the scales or something.

👍

tedmoore:
> There is only one section missing in whyscale is the importance of log transform at times.

For sure. This is a really good addition.

PR #320, opened 2022-05-25
RobustScale

tedmoore:


tremblap:
this is good.
a quick sentence I needed to read many times: 
"a range of 1 between two specified [percentiles]"
the word range confuses me out. it might be a 2nd language thing.

what about:

Robust-Scaled data means that each dimension in the scaled data meets two criteria:
1. its median will now be 0
2. its values will be scaled to insure that the range between 0 and 1 will be equivalent to the distance between two specified [percentiles](these are often the 25th and 75th percentiles).

(I know it is a bit wordy but I presume the english language can be improved :) what matters here is the idea of transform, a then and now narrative)

tremblap:
I would add a sentence about IQR, feathered bar graph, and the link to the Khan academy, at the bottom, so the eager reader is channeled there.

(I really like the comparing scaler subpage)

src/routes/reference/robustscale.svx
```diff
@@ -1,11 +1,34 @@
...
+</script>
+
+Robust-Scaled data means that each dimension in the data meets two criteria:
+
+1. it has a _median_ of 0 and 
+2. a range of 1 between two specified [percentiles](/reference/bufstats#order-statistics) (these are often the 25th and 75th percentiles). 
+
+RobustScale transforms a [DataSet](/reference/dataset) so it meets the above criteria by analysing the DataSet for the appropriate statistics and scaling each dimension accordingly.
+
+Because RobustScale is based on percentiles, the scaling of the data is less affected by extreme [outliers](/learn/outliers) than other scalers (such as [Standardize](/reference/standardize) and [Normalize](/reference/normalize)). Being more robust to outliers can help RobustScale more accurately scale and represent the majority of the data.
```
jamesb93, line 24: This is a valuable link to point to in this regard.

https://scikit-learn.org/stable/auto_examples/preprocessing/plot_all_scaling.html

jamesb93 (changes requested):
I feel that the opening sentence is weird and needs to be restructured so that there is an introduction more similar to other references. It starts too specific in other words.

I also added a reference that I think is worth pointing at.

tedmoore:
> the link to the Khan academy

Can you supply this link?

PR #323, opened 2022-05-30
AmpGate

tedmoore:


jamesb93 (approved):
I like the graphic and you've done a good job trying to cram all the different constraints in there. Again, I wonder if an interactive thing would serve us well here. Imagine you can adjust all the thresholds and see lookback/lookforward in realtime. 

I did have that realtime version many moons ago that was doing a very minimal envelope follower but that was mostly limited by the fact that it was realtime. We could make something really good in the offline space together with me driving the HTML/JS code and you driving the concepts.

PR #357, opened 2022-07-28
KNNRegressor Reference

tedmoore:


jamesb93:
is there a visualisation we could do here to help people understand what the KNNRegressor does? Happy to code something up if someone else can imagine it @tedmoore 🙊

tedmoore:
I think the idea right now is to go light on this one until/if the multi-to-multi implementation get's done. At which point this can be fleshed out/changed to align with that functionality. That's my understanding from @tremblap.

tremblap (changes requested):
that is a good place holder until I make it HD :) There is an ommission: the return is either a weighted or simple average between the K points, depending on the parameter weight (IIRC)

tedmoore:
> that is a good place holder until I make it HD :) There is an ommission: the return is either a weighted or simple average between the K points, depending on the parameter weight (IIRC)

good catch!

PR #358, opened 2022-07-28
NMFMorph Reference

tedmoore:


tremblap (approved):
this is terse but good - I'm surprised you didn't explain in a one liner what optimal transport is though, as you seem to enjoy doing these sort of explanations and do them well, but there is a paper, so all is fine.

tedmoore:
> this is terse but good - I'm surprised you didn't explain in a one liner what optimal transport is though, as you seem to enjoy doing these sort of explanations and do them well, but there is a paper, so all is fine.

This is a good point. It might be something that would be nice to have as a separate learn article. That way both NMFMorph and AudioTransport could link to it (are there other Fluid objects that use it too? I think that's the only ones right?)

PR #359, opened 2022-07-29
Transients Reference

tedmoore:
I have an impulse that this could have a diagram similar to https://learn.flucoma.org/reference/ampgate/ampgate-01.jpg however that diagram is already quite/too busy (it's _as busy_ as the object's parameters are). And Transients is an even more complicated algorithm, so I'm not sure that more arrows is going to help (or be inviting to) a user in this case. 

If anyone has an image in their mind of what this could look like and be the right amount of information/elucidation that would be useful. 

If not, I think this degree of explanation (in coordination with the help files) is useful and approachable. Thoughts?


tedmoore:
This is `build`ing fine on my machine, so maybe @jamesb93 can test it locally and see what my problem is? Maybe still something to do with me needing to re-install node?

jamesb93:
> This is `build`ing fine on my machine, so maybe @jamesb93 can test it locally and see what my problem is? Maybe still something to do with me needing to re-install node?

Seems fine, if you merge it I think it will work 99.9%

tedmoore:
👍 🙏

tremblap (approved):
clear again - the formula made me think for a second but hey, that makes sense - it is just backwards of what I thought it was doing without thinking...

PR #367, opened 2022-08-08
suggested changes from moore

tedmoore:
A few very minor suggestions. They're all for accuracy where I could make a light touch. The main two are that:

1. the title is with a lower-case "q"
2. that document is not _technically_ my dissertation, a more accurate description for it would be my PhD research.

let me know if you have any questions about any of it!


jdchart:
This all looks great! Merged, and will be pushing the article tomorrow!

tedmoore:
❤️ 🙏

PR #368, opened 2022-08-09
BufNMFSeed reference

tedmoore:


tremblap (approved):
wow I understand it now - this is great thanks!

PR #372, opened 2022-08-11
[content] bufcompose, nov/amp/onset feature

jamesb93:
Adds reference entries for:

- NoveltyFeature
- AmpFeature
- OnsetFeature
- BufCompose

Would like some quick feedback if you have time @tedmoore.


tedmoore (changes requested):
Super. I have a few thoughts:

**BufCompose**

* I think the subtitle for it can be less tautological than "Buffer composing". Maybe "Copying contiguous data from one buffer to another"
* `targetted` is a typo
* for startFrame and numFrame, the default is actually 0, not -1
* I would prefer to have the arguments in the order that they appear in in the object, so startFrame, numFrames, startChan, numChans.

**NoveltyFeature**

* In the pointer box at the bottom, I think the "the feature object" should be changed to "NoveltyFeature" and "the companion slicing object" should be changed to something like "its companion slicing object, NoveltySlice" to make it more clear. I also think this info should be like the second sentence on the page! 

**OnsetFeature**

* Same about the pointer box

**AmpFeature**

* Same about the pointer box

PR #380, opened 2022-08-18
Getting started formating

jdchart:


jdchart:
Top bit is better, now with cards. Except the flair which looks a bit weak, especially when screnne smaller.
Other stuff down below should prpobably actualy be tiled with two items per line cause there's so much stuff.

Issue #396, opened 2022-12-22
Incorrect URL for the video


AkiyukiOkayasu:
The URL to the video listed in [Classifying Sounds Using a Neural Network](https://learn.flucoma.org/learn/classification-neural-network/) is incorrect.
Only "Max" is wrong and "SuperCollider" is correct.

Specifically, the URL below is wrong, and the correct link is [this](https://youtu.be/cjk9oHw7PQg).
https://github.com/flucoma/learn-website/blob/545b1856c147a466395ad9c01ad2df4f2fa276d1/src/routes/learn/classification-neural-network/%2Bpage.svx#L35

AkiyukiOkayasu:
The URL for [Controlling a Synth using a Neural Network](https://learn.flucoma.org/learn/regression-neural-network/) is also incorrect as well.  
The Max video is also incorrect.

PR #399, opened 2023-01-11
Fix URL for "Classifying Sounds with a Neural Network in Max" and "Controlling a Synth using a Neural Network in Max"

AkiyukiOkayasu:
Fixed issue https://github.com/flucoma/learn-website/issues/396


jamesb93:
Thanks for the PR! I'm having a look into this shortly.

jamesb93:
Thanks for this. I just wanted to check it compiled and there were no issues merging before. This is really helpful!

PR #405, opened 2023-03-02
Design Improvements

jamesb93:
This PR adds a significant design overhaul to the learn platform while still retaining the general aesthetic and visual language as before.

Major changes include:

- Accessibility with navigation by adding keypress to div(s) that appear as <a>
- Squareification of buttons and card-like components
- More subtle flagging of flaired content (article, events, etc.)
- Header is now sticky and reworked as a "global" navigation

Amongst this big changes are subtle ones such as more complementary and readable font sizes, minimisation of <div> nesting and overall cleanup of CSS.


jamesb93:
To check this:

1. Checkout the PR
2. `npm run dev`
3. Go to `localhost:5173`
4. Look at the site

tremblap (changes requested):
Fantastic look improvement. One quick mod proposal: when we get to the hamburger menu, is there a way to keep the flucoma logo on the top left? (without the linking "FluCoMa Learn" words, I don't care that much about them then). I just miss branding a bit - the rest is soooo slick. For instance, in halfscreen in HD, there is no trace of FluCoMa at all in the header (as per the attached pic)
Screenshot 2023-03-03 at 09 48 41

PR #407, opened 2023-03-09
RSS Syndication for Explore Podcasts

jamesb93:
This adds RSS syndication by providing a server endpoint that aggregates podcast markdown and its associated metadata into the RSS XML format.

- add rss endpoint for explore
- correct gborn typo
- make the wrapper a div and the content a main
- make flairs contrast better with background
- update YouTube chapters styling for RSS


jamesb93:
This is non-breaking. So I'm going to merge it to start testing :)

PR #410, opened 2023-05-08
attempt at adding the link

tremblap:
after a few users complained they couldn't find a link to installations, I tried to add it to the home page of 'getting started'

I strangly cannot use the links at that level, and I wonder why.  help?


jamesb93:
The reason you cannot use markdown syntax in this file is because it is a `.svelte` file. Unfortunately you'll have to hand write the HTML :)

I pushed a commit which addresses this and you can compare for your own knowledge

jamesb93 (approved):

PR #427, opened 2023-08-28
Learn site typos

MattS6464:
A correction of typos/grammar errors across the learn and reference pages on the learn website.


jamesb93:
This is great and a significant number of these were me 👉👈.

Did you catch these with a script or just manually? If it was a script we should integrate it into a pre-flight check.

MattS6464:
I caught them manually :)

PR #428, opened 2023-08-28
Where is my data tutorial

MattS6464:
Added the where is my data patch (padding tutorial) into the learn website examples for both Max and PureData.


MattS6464:
Fixed a typo in multiple places saying there was 1000 samples of a sine wave when it should say 1024.

PR #430, opened 2023-09-03
Fix MFCC ref page "Drop Zero" buttons wrong way round

fearn-e:
MFCC reference page's drop zero example has the "startCoeff = 0" and "startCoeff = 1" buttons the wrong way round, they do the opposite of what they say.


jamesb93:
Good fix! Simple but an egregious one from my part.

PR #431, opened 2023-09-03
Fix MFCC ref page "Drop Zero" analyser not always starting with audio playback

fearn-e:
MFCC reference page's drop zero example doesn't start analyser with audio playback after the first time it's paused, requiring a page refresh. Seems like the `analyser.start();` line is simply placed one curly bracket too early.


jamesb93:
Another good catch. Thanks!

PR #453, opened 2024-10-17
navigation to/from the podcast page

tremblap:
adding a return link to the podcast, and reordering the tags of the landing page so they are in the same order as the menu bar

this is a bit crude but the podcast page layout is not the same and there was no way to come back. Ideally it would have the same header than the rest of learn, with all the menus, but I didn't manage to do that...


tremblap:
I also changed the order of the main page's buttons to fit the many order

jamesb93 (approved):

PR #455, opened 2025-01-15
Added missing SC "starter file" for classification activity

tedmoore:
This file seems to be missing. I found a link *to* it but got a 404 so I just made one and added it in...

@jamesb93 before I merge this, just want to check that the automated deployment is still in place? Or is there a new workflow I am un aware of?


jamesb93:
It's been a hot minute since I looked at it but it should be working. I believe PA successfully deployed in December or so. Whether or not that was automatic, or a button had to be pressed I can't remember...

Try merging your PR and then seeing if the changes are reflected.

jamesb93 (approved):

PR #462, opened 2026-05-22
Fix/discrepancy get started

tremblap:
An astute web user pointed at me that we were using Getting Started and Get Started inconsistently within our site. This corrects that. Asking mature web designer for opinions... and if we should put a simlink from the old addresses (https://learn.flucoma.org/learn/getting-started/ and https://learn.flucoma.org/getting-started/) to the new ones (https://learn.flucoma.org/learn/get-started/ and https://learn.flucoma.org/get-started/)

The other option is the first commit of this PR, which changes only the UX side, but not the structure/link of the site.

let the vote begin :)


tedmoore (approved):

jamesb93:
I would change the surface but not the underlying link. That way we can depend on any existing URLs working. I suspect the CCE packages link to it, so worth keeping the same IMO.

tremblap:
ah shit. I had an epifani in holidays that I remembered the push was wrong because of this. @jamesb93 2 questions: should we have both links cohabitating (the corrected one and the old one as forward) to make it less ugly? or just keep the old site syntax with the new 'surface' as you call it?

tremblap:
the good news is that my PR was in 2 steps: one for the 'surface' then one with the file-naming. Anyway, let's think on how to make it consistant yet backward compatible (and then we can fix your current PR with the solution we pick)

tremblap:
maybe the good way round is to merge back to my middle commit (faces not files) and do an attempt PR at moving files where they belong (my 2nd commit) but including automagically making forward links (which i would need claude help to find a way to do in our framework)

tremblap:
(that would allow @jamesb93 to merge his fixes I think, or at least with very small changes)

